PR #23, opened 2015-08-13
Small fixes for Visual Studio

libertyernie:
These are all fairly minor fixes that help get the project compiling in Visual Studio 2013.

I haven't gotten the AAC and MP4 libraries to work yet, so to compile in VS2013 I took out the defines that enabled them in the project settings (not included in this pull request.)


libertyernie:
By the way, I got MP4 and AAC support to compile using your other repos, so that's good now.

PR #24, opened 2015-08-30
Use built-in snprintf on VS2015

libertyernie:
Visual Studio 2015 now has snprintf defined.


kode54:
Please note that version.bat is still useful for building Windows binaries. You need only invoke it with a full path to the header file it will generate, and the name of the macro it will generate, and it will produce the same tag as version.sh. Since I found no convenient way to make Visual Studio simply pipe the output of a script into a header file, I chose instead to generate it myself in a pre-build step.

kode54:
Also, I mostly track this project on my own Gitlab repository now, but I will try to merge and incorporate changes from here when I can, and push my changes here, since people are still forking it.

libertyernie:
OK. The version.bat I got rid of was someone else's (from another fork) and it wasn't working for me. If I do any more work in my fork I'll put yours back in.

kode54:
Mine currently is in the pre-build step in the foobar2000 and xmp-vgmstream vcxproj files. It would need to be added to the .vcproj files as well, if possible. It uses extended batch processing features for things like piping command output to a variable using the for command, and simple string search and replace. I did update it in a new commit to change the git describe command to match version.sh, and it appears to work properly.

PR #43, opened 2016-11-27
Decryption keyfiles

bnnm:
Add a way to load decryption keys from external files, currently for ADX.

Needed for some songs where vgmstream guesses the wrong key (Gunhound PSP).
Also handy to try new keys without having to recompile.

The format is "(songname.ext)key" (per song), "(.ext)key" (per folder)


bnnm:
Thanks, makes sense. Updated.

The filename string stuff is kinda messy too :(

PR #44, opened 2016-11-27
FFmpeg fixes

bnnm:
Minor ffmpeg tweaks and added AT3/MP3 via FFMpeg to PS3 MSFs. MP3s have some crashing issues I'll fix soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-.svn
+.*
+!/.gitignore
 *.user
 *.suo
 *.vspscc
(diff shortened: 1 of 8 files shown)


kode54:
I'll wait on this for you to deal with the crashing issue? I hope you can fix it in short order. Good luck.

bnnm:
Sure, I got it mostly fixed. I thought it could be easier for you to review a separate pull request, since it changes a few core FFmpeg parts.

The main issue is that FFmpeg demuxers (like MP3s) apparently don't need to set "duration" (since FFmpeg handles raw/unknown/live streams), but vgmstream FFmpeg code expects it, so it does funny things without it.

I'm tweaking it so its less dependant on "duration" (totalFrames/framesRead/etc).
Or I could externally feed it the duration, but I think it's better to have it decoupled.

PR #45, opened 2016-11-29
FFmpeg: keep block size constant (for edge loops), don't write on EOF

bnnm:
Fixes a couple of crash bugs.
- Changing the block size dynamically was messing with vgmstream's own stuff (ex, loops from num_samples to sample 0 would have totalFrames-framesRead=0 when looping)
- Added a few extra checks when duration/totalFrames isn't set.

Unfortunately there is still a subtle looping bug. Somehow it drops the next decoded frame after looping (it happens even if you manually seek to 0 avformat_seek_file at the start even before decoding anything).
All I checked so far seems fine (AVIO buffer reading/seeking offset, frame reading/decoding, different formats...), not sure what else could be wrong.

### DIFF
--- a/src/coding/ffmpeg_decoder.c
+++ b/src/coding/ffmpeg_decoder.c
@@ -90,7 +90,7 @@ void decode_ffmpeg(VGMSTREAM *vgmstream,
     
     int framesReadNow;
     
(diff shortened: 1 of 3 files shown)


kode54:
Awaiting further changes before I attempt to produce a build with FFmpeg support for Windows.

PR #46, opened 2016-12-03
Ffmpeg fixes4

bnnm:
- Fixed some formats not properly seeking to 0 and validate they can (see ffmpeg.c init_seek)
- Fixed discard in edge cases, it should always work now
- Don't check framesRead and rely on EOFs (easier to fix bad headers out there)
- Cleanups and minor fixes

With ACCURATE_LOOPING flag it seems correct now, and I hope to fix regular looping later.
Frustratingly some odd format (.shn) still won't seek at all, even with manual avio_seeks. I'll persevere though.

There are further fixes/formats/tests I want to do, if you want to wait on doing a build with FFmpeg.
I'll notify when I feel it's stable enough.


kode54:
Continuing to watch this progress eagerly.

PR #59, opened 2017-01-08
Formats: VAWX , VDS/VDM, CXS; description lists

bnnm:
- Added .nps extension [Venus & Braves PS2]
- Added ATRAC3/XMA in VAWX (.xwv) [Moon Diver, No More Heroes X360]
- Added .swag [Frantix PSP]
- Added PS2 VDS/VDM [Graffiti Kingdom]
- Added CXS [Eternal Sonata (Xbox 360)]
- (dev) Moved coding/layout/meta switch-case descriptions to list in formats.c
- (dev) Minor cleanups
- (dev) FFmpeg: remove ACCURATE_LOOPING flag now it's battle-tested

No separate branch because I forgot.


src/meta/ps2_npsf.c
```diff
@@ -14,7 +15,9 @@ VGMSTREAM * init_vgmstream_ps2_npsf(STREAMFILE *streamFile) {
 
     /* check extension, case insensitive */
     streamFile->get_name(streamFile,filename,sizeof(filename));
-    if (strcasecmp("npsf",filename_extension(filename))) goto fail;
+    if (strcasecmp("npsf",filename_extension(filename)) &&
```
kode54: Shouldn't this also use header_check_extensions?

bnnm: Well. Since I'm sort of self-evaluating the viability of header.c this one was a quick update.

I'm not super-confident adding a new "main" src file into a venerable structure, which in the end only I will use. I'm considering just moving its funcs into streamfile.c/util.c/vgmstream.c for simplicity.

formats.c I'm quite satisfied tho, couldn't accept all those gigantic if/switch-case lists.

PR #72, opened 2017-03-05
Stuff

bnnm:
- Fixed some XWB XACT2 looping [Blue Dragon, Prey]
- Added basic multi-stream FSB5 support
- Added SFX SGDX support
- Shows stream numbers in the description (few multi-stream formats only)

### DIFF
--- a/src/meta/ps3_sgh_sgb.c
+++ b/src/meta/ps3_sgh_sgb.c
@@ -28,14 +28,14 @@ VGMSTREAM * init_vgmstream_ps3_sgdx(STREAMFILE *streamFile) {
     VGMSTREAM * vgmstream = NULL;
     STREAMFILE * streamHeader = NULL;
 
-    off_t start_offset, data_offset;
-    int32_t data_size;
+    off_t start_offset, data_offset, chunk_offset;
+    size_t data_size;
 
     int is_sgx, is_sgb;
     int loop_flag, channels, type;
     int sample_rate, num_samples, loop_start_sample, loop_end_sample;
 
-    int target_stream = 0;
+    int target_stream = 0, total_streams;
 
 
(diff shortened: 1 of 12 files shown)


src/meta/ps3_sgh_sgb.c
```diff
@@ -70,24 +70,23 @@ VGMSTREAM * init_vgmstream_ps3_sgdx(STREAMFILE *streamFile) {
...
+    if (is_sgx) { /* position after chunk+size */
+        if (read_32bitBE(0x10,streamHeader) != 0x57415645) goto fail;  /* "WAVE" */
+        chunk_offset = 0x18;
+    } else {
+        if (!find_chunk_le(streamHeader, 0x57415645,0x10,0, &chunk_offset,NULL)) goto fail; /* "WAVE" */
     }
 
+    /* check multi-streams (usually only SE containers; Puppeteer) */
+    total_streams = read_32bitLE(chunk_offset+0x04,streamHeader);
+    if (target_stream == 0) target_stream = 1;
```
kode54: Huh?

bnnm: That's, uhm, a bit vague.

If you ask about the chunk_offset, it points after the id+size = 4+4 (makes more sense to me; say you have find_weird_chunk with id+size = 2+variable, you want it to return an offset to the chunk data, after the variable size)

If you ask about total/target_stream, if I implement a stream selector for multi-stream formats later, I'd think 1=first stream, N=Nth, while 0=none selected. Which now it means first stream, but I could make it give an error when none selected.

Otherwise, all sample files I have decode the same.

kode54:
Oh, sorry. I thought I saw the code also range checking the stream selector like a 0-based index, which confused me a bit.

PR #87, opened 2017-05-01
Plugin cleanup

bnnm:
- (dev) Updated Audacious plugin and added build instructions
- (dev) Plugin and other cleanup, renamed stuff for consistency

I've renamed some plugin files and the audacious/xmplay folders because it bothered me so much.
But I can revert them if you prefer, no problem.
(I was tempted to rename fb2k to foobar2000 too, but maybe it would mess with your config).


FraGag:
I tried the updated Audacious plugin with Audacious 3.8.2 on Arch Linux. When I try to play a file supported by vgmstream, the file fails to start playing and Audacious gives an error: "Erreur de lecture des métadonnées" (which is French for "Error reading metadata").

After cleaning the worktree, I compiled the plugin with:

```
$ ./bootstrap
$ ./configure
$ make -f Makefile.audacious
$ sudo make -f Makefile.audacious install
```

kode54:
Please check this pull request, then file a fresh bug report if it still fails.

PR #102, opened 2017-06-23
High Disk usage when playing BFSTM files on endless loop

faxx1080:
Noticed that for certain BFSTM files (Breath of the Wild), there was intensive disk usage (9 GB/s reported by process hacker) when playing in an infinite loop. This restores old behavior which fixes this issue.


kode54:
How is that even possible? Consumer SSDs read in the sub-GB/s speeds, while HDs read around 120MB/s.

faxx1080:
If I run the current master build of test.exe on a 6 channel BFSTM file, I get this:

result-bgmcastle

But if I run it on a 2.2 MB stereo file, I instead get what seems to be infinite recursion:

result-sheiktower

As for why the 9 GB/s, I think the decode_ngc_dsp function is called *very* fast, potentially causing a resource leak? I haven't looked that far yet. For what it's worth, the real read speed on my laptop is ~1.8 GB last I checked.

kode54:
And I am telling you, there is no way that your drive is reading at 9GB/s. That is most likely cache hits. And even then, the way you're using it is incorrect. The test process was not designed for piping IO into a WAV file player, so you're probably using it unthrottled as the pipe buffer continues to grow. Have you verified whether actual player plugins read from your drive at "9.5GB/s" somehow? You do know, if it were reading that much data per second, your drive would probably slow down the playback.

faxx1080:
I see that now. When I use the r1050-536-gf670ae0 version (on the foobar2000 download page) I see what you mean - task manager reports normal disk activity, but Process Hacker is still saying around 300 MB/s. My question in that case: is that expected behavior?

kode54:
It may be, if the audio is being piped to an external process. Are you using either ASIO or WASAPI with foobar2000? It is also quite possible that the test.exe's pipe output was being counted into that field. As for stereo files recursing infinitely, that sounds like a bug.

bnnm:
Due to how cache/reading works it's possible for certain files to trash the buffers and trigger continuous fseeks/freads in decode_ngc_dsp (or any other decoder), which I assume is your case.

Probably a bug/something to improve in vgmstream_open_stream (main diff in the commit) with 6ch files, as it's a generic handler to set up buffers. I'll check with that game and report.

That said I'd assume the OS and or HDD would also cache those reads, so I don't think it's actually transfering that much.

bnnm:
I've pushed a change that improves buffers in multichannel files, so it should work now as before for BFSTM.
Thus I don't think this PR is needed anymore.

PR #104, opened 2017-07-01
EA SCHl, MED

bnnm:
- fixed some .MED with JUNK chunk
- redone EA SCHl support [EA games from roughly 1997~2010]
  * tested most variations using EA's tools
  * added looping [Need for Speed II PC]
  * added DSP support [3DS games]
  * added MP2 (tested) and MP3 (untested) [Harry Potter and the Chamber of Secrets (PS2)]
  * fixed PCM/EAXA/MT in some cases [PS2/GC/XBOX/Wii games]
  * fixed sample rate/codec defaults, num_samples in some cases
  * documented more header patches and other details (ex. "EA ADPCM" is now EA MT/MicroTalk, as per EA's tools)
  * (dev) some internals adapted for EA stuff


kode54:
Doesn't fix the .eam files I ripped from Need for Speed II PC demo. They appear to include external files which map the blocks, as well.

bnnm:
I'll take a look, can you post some samples? The NFS2 PC .EAM from joshw seems to work fine though.

Support for those MAP+EAM is a bit hacky but should be enough, I take the .MAP only has pointer to offsets.

kode54:
SHOW.eam, the main menu music, plays until it hits the end, then freezes or something. Maybe it's not supposed to be the correct file?

bnnm:
Hmm. Maybe we have different versions? Could you post the file? Here is what I have:
- NFS2 full: found at joshw, has SHOW.EAM, crc 7AF708AC
- NFS2 Demo (nfs2demo.exe ~16MB): contains PACROCK.ASF and SHOW.MUS, crc 22EB62B1.
- NFS2 SE Demo (Nfs2sea.exe ~6MB): only has LASTROCK.ASF, and uses the interleaved MT I mentioned.

Both SHOW are ~5.43MB, 3:58s / 5251829 samples and play fine for me (they are mastered differently though).
The song seems to end kind of abruptly but I checked and that's how it's recorded, also matches youtube vids. Does it manually loop in-game or such?

kode54:
The .map has indications to loop it, from the last block/range to the second block/range. It's not the smoothest loop, either. Maybe the game has hard coded loops?

bnnm:
Ah, in that case I'll try to extract loops from the .MAP for the next revisions. I see MAP are supported in foo_adpcm so that'll save some time.

kode54:
Another thing. I have the player, Cog, which is designed to play until the vgmstream library signals EOF, or until 2 loops and a fade if loops are enabled, and for SHOW.eam, it continues playing silence after the file cuts off.

bnnm:
I've tweaked a bit the EA block handling on EOF in the last PR, hopefully fixes Cog.

PR #117, opened 2017-08-13
Streams, plugins

bnnm:
- (dev) plugin cleanup in preparation of future changes
- Make substream selectable for some formats
- Read internal filename for some formats
- test.exe: Add stream selector flag "-s N"
- winamp: add unicode support (for names/dual files)
- Redo Angel STM: add PCM, adjust looping, unify code, add .lstm [Red Dead Revolver, Spy Hunter 2]

You may want to check foobar's code in case I did something dumb. I prepared/moved stuff for subsong handling and easier understanding. It's working ok here but still under testing so it's disabled.


kode54:
Never mind, I'll make that change myself.

PR #126, opened 2017-09-24
FFmpeg custom, Opus

bnnm:
- Fix some XNB [Dragon's Blade (Android)] + .DA [The Great Battle VI (PS)]
- Add Wwise Opus [Mario + Rabbids Kingdom Battle (Switch)]
- Add EA-XMA for EA SNU [Dante's Inferno (X360)]
  * only 1/2ch works correctly as multichannel layout need to be investigated
- (dev) Add FFmpeg custom IO callbacks for data transmogrifiers
- Ogg Opus support (renamed to ~.logg or such) [Ken Follett's The Pillars of the Earth (PC)]
- Add some HCA keys

FFmpeg custom IO may be a bit hard to follow, but I preferred this to making frame-by-frame FFmpeg or libopus decoders. Probably could be implemented with STREAMFILEs instead, seemed too complex for me though. Previously supported FFmpeg formats should work the same.

for FFmpeg Opus (Wwise, Ogg) please add:
* --enable-parser=opus
* --enable-decoder=opus
and remove:
* --disable-swresample

this means adding a new .DLL and friends I'm afraid:
* ext_libs/swresample-vgmstream-2.dll
* ext_libs/swresample-vgmstream-2.dll.asc
* ext_libs/swresample-vgmstream-2.def
* ext_includes/libswresample/swresample.h
* ext_includes/libswresample/version.h

I changed Visual Studio's project files too but I could only try test.exe/winamp ATM, in case you have trouble building it.


kode54:
Why is swresample even needed? Are there cases where we want Opus to output something other than 48kHz?

bnnm:
Yeah, that one got me stumped for a while, dependency being buried in the [configure](https://github.com/FFmpeg/FFmpeg/blob/master/configure) script and all.

PR #170, opened 2018-01-07
Makefiles, etc

bnnm:
- Add HCA key
- (dev) Update Makefiles, docs, etc


ext_libs/Makefile
```diff
@@ -30,7 +36,9 @@ libswresample.a: swresample-vgmstream-3.dll swresample-vgmstream-3.def
 	$(DLLTOOL) -D swresample-vgmstream-3.dll -d swresample-vgmstream-3.def -l libswresample.a
 
 libatrac9.a: libatrac9.dll libatrac9.def
-	$(DLLTOOL) -D libatrac9.dll -d libatrac9.ef -l libatrac9.a
+	$(DLLTOOL) -d libatrac9.dll -d libatrac9.def -l libatrac9.a
```
kode54: Shouldn't that first switch be a capital D?

Thealexbarney:
Is there a makefile for mingw builds from Linux now?

Also continuing from [here](https://hcs64.com/mboard/forum.php?showthread=52892&showpage=3#post_53874), did @kode54 decide on whether to continue dynamically linking libatrac9, or to include the source directly in vgmstream like the HCA library source is?

ext_libs/Makefile
bnnm: Why yes, it should (leftover from a prev change; I think -D isn't needed if the .dll name is included in the .def though)

bnnm:
@Thealexbarney - I believe the Makefile.autotools.am now builds test.exe (as vgmstream_cli for Linux), vgmstream123 and Audacious, should include mpg123/vorbis.

I *think* the regular Makefile now works if you pass/change TARGET_OS=LINUX (didn't test yet as it takes me a while to setup Linux), all external libs are disabled though, maybe could be supported with .so?

PR #180, opened 2018-01-20
Automate MSVC builds, and setup continuous integration

Thealexbarney:
- Add a script to download dependencies
- Move dependencies to "/dependencies"
- Update solution file so it contains all projects
- Add a script for building and packaging the MSVC builds
- Add information for using AppVeyor's continuous integration service

Changes to the projects themselves are minimal. I only needed to change the dependency paths, and update the solution configurations because it didn't include all the projects.

All that's needed to enable CI builds is adding the project to AppVeyor Names and version strings can be adjusted if necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,7 @@ Release
 /xmplay/Debug
 /xmplay/Release
 /xmplay/*.dll
+/dependencies
+
+/**/test.zip
+/**/foo_input_vgmstream.fb2k-component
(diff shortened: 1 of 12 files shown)


Thealexbarney:
Since it's not running on your repo, here's an example of this PR on AppVeyor:
https://ci.appveyor.com/project/Thealexbarney/vgmstream/build/r1050-950-g5c8f4bb.PR3.2

PR #186, opened 2018-01-28
A small fix for Android. 

haspor:
The function in streamfile.c line 164 (open_stdio) was behaving strange on Android, most likely due to those defines in streamfile.h starting at line 19.

Songs from this set fails to play without the fix: Bionicle Heroes (2006-11-14)(Traveller's Tales)(Eidos) and possibly some others too.


kode54:
I guess we never expected anyone to be using MSVC to build for Android.

PR #187, opened 2018-01-30
Fixes for Switch Opus reading

Thealexbarney:
- Fixes getting the sample count from Nintendo Switch Opus streams. See the commit message for details.
- Read the pre-sample count from Switch Opus headers.

~~Try and make sure the skip samples are set when using ffmpeg. The value is almost always put in initial_padding.~~

### DIFF
--- a/src/coding/ffmpeg_decoder_utils_switch_opus.c
+++ b/src/coding/ffmpeg_decoder_utils_switch_opus.c
@@ -169,7 +169,7 @@ size_t switch_opus_get_samples(off_t offset, size_t data_size, int sample_rate, 
         uint8_t buf[4];
         size_t block_size = read_32bitBE(offset, streamFile);
 
-        read_streamfile(buf, offset+4, 4, streamFile);
+        read_streamfile(buf, offset+8, 4, streamFile);
         num_samples += get_opus_samples_per_frame(buf, sample_rate);
 
         offset += 0x08 + block_size;
--- a/src/meta/nsw_opus.c
+++ b/src/meta/nsw_opus.c
@@ -9,7 +9,7 @@ VGMSTREAM * init_vgmstream_nsw_opus(STREAMFILE *streamFile) {
     int loop_flag = 0, channel_count;
     int num_samples = 0, loop_start = 0, loop_end = 0;
     off_t offset = 0, data_offset;
-    size_t data_size;
+    size_t data_size, skip = 0;
 
     /* check extension, case insensitive */
     if ( !check_extensions(streamFile,"opus,lopus,nop")) /* no relation to Ogg Opus */
@@ -53,6 +53,7 @@ VGMSTREAM * init_vgmstream_nsw_opus(STREAMFILE *streamFile) {
     channel_count = read_8bit(offset + 0x09, streamFile);
     /* 0x0a: packet size if CBR, 0 if VBR */
     data_offset = offset + read_32bitLE(offset + 0x10, streamFile);
+    skip = read_32bitLE(offset + 0x1c, streamFile);
 
     if ((uint32_t)read_32bitLE(data_offset, streamFile) != 0x80000004)
         goto fail;
@@ -77,10 +78,9 @@ VGMSTREAM * init_vgmstream_nsw_opus(STREAMFILE *streamFile) {
 #ifdef VGM_USE_FFMPEG
     {
         uint8_t buf[0x100];
-        size_t bytes, skip;
+        size_t bytes;
         ffmpeg_custom_config cfg;
-
-        skip = 0; //todo
+        ffmpeg_codec_data *ffmpeg_data;
 
         bytes = ffmpeg_make_opus_header(buf,0x100, vgmstream->channels, skip, vgmstream->sample_rate);
         if (bytes <= 0) goto fail;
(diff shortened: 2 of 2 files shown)


bnnm:
@Thealexbarney - Been a while since I checked but I don't think the changes at ffmpeg_decoder.c@558 are correct.

IIRC when FFmpeg sets skip_samples/initial_padding/whatever, it'll internally skip samples. If you call ffmpeg_set_skip_samples, it'll  hackily override that, which I don't think we really want, since they use several skip fields there is a chance it'll incorrectly re-skip.

It sets data->skipSamples so metas can check if it they should manually call ffmpeg_set_skip_samples().
For SwOpus I think skip should work without changes if passed to ffmpeg_make_opus_header.

Hope that makes sense.

Thealexbarney:
I wasn't able to find any instances where codecs using FFmpeg would correctly skip samples with the current master branch. Formats like Ogg Opus, and ATRAC3 via FFmpeg still have their initial padding when decoded.

bnnm:
I tested ATRAC3 and skips fine for me. I think AAC worked fine too.

If Opus skip isn't working either make_opus_header and co is making something wrong, or maybe the FFmpeg DLL isn't actually skipping. If it's the latter I designed ffmpeg_set_skip_samples to be used externally (something like ex. riff.c@480 in nsw_opus.c), because if skip gets fixed in the DLL it should let it do it rather than forcing vgmstream's skip hack.
Given how many quirks FFmpeg has (ie. see the grotesque init_seek or how they have a bunch of similar-looking skip fields in their code) that's the way I thought would be safer.
Either way it's something to test carefully.

Thealexbarney:
Did you make sure to disable `maiatrac3plus` so that ffmpeg would be used to decode the ATRAC3 file?

It seems like codecs which set `initial_padding` in the codec context don't skip any samples, but those that set `skip_samples` in AVStream do, based on my limited testing.

I don't know if this is something with the way vgmstream is using libav, or if that's how it's supposed to work.

kode54:
I'll be disabling maiatrac3plus in the next release anyway, if I manage to remember to do so. So that won't be an issue, hopefully.

Thealexbarney:
@kode54 It was actually ffmpeg that didn't skip samples, and maiatrac3plus that did

bnnm:
Actually it was the other way around, guess I wasn't too clear.

Skip recap:
-ATRAC3plus w/ maiat3p: no skip, complex to fix
-‎ ATRAC3/plus w/ FFmpeg: needs manual set_skip in meta ATM, if fixed in FFmpeg DLL set_skip would be ignored
-‎ XMA : no skip, must not manually use set_skip due to bugs in FFmpeg DLL
-‎ MPEG: not tested but I think FFmpeg DLL skips them if LAME header/etc is found.
-‎ AAC/AC3 (raw): must set_skip in metas (no internal info)
-‎ MP4+AAC: FFmpeg DLL skips IIRC
-‎ OGG Vorbis: not skipped ATM due to init_seek problems, to be fixed
-‎ OGG Opus: didn't test yet but I think FFmpeg DLL doesn't

I felt forcing the use of set_skip in ffmpeg_decoder.c gave unpredictable results, which is why I deliberately only call it in select cases.

If initial_padding is ignored like you say and you force set_skip, there is a chance samples are skipped twice if they change internals to honor it. So I think this should be fixed in other way identifying the actual issue.

Thealexbarney:
Here are my results of whether any samples were skipped when decoding with vgmstream+ffmpeg, along with the variables that were set with the number of skip samples after opening the file. 

It's even messier than I initially thought it was. Command line ffmpeg behaved the same way as vgmstream+ffmpeg did in regards to skipping samples.

```
Ogg Vorbis:
Skips
initial_padding: 0
start_skip_samples: 0
skip_samples: 0

Ogg Opus:
No skip
initial_padding: Set
start_skip_samples: 0
skip_samples: 0

Atrac3+:
No skip
initial_padding: 0
start_skip_samples: 0
skip_samples: 0

aac m4a:
Skip
initial_padding: 0
start_skip_samples: 0
skip_samples: Set

mp3:
Skip
initial_padding: 0
start_skip_samples: Set
skip_samples: 0
```

bnnm:
Some FFmpeg formats skip using negative packet timestamps (ex. Vorbis), but not all negative timestamps mean skip samples I believe. IIRC there was another way to skip (flags?) too. That's why I insist on calling ffmpeg_set_skip_samples manually in metas, it's way too fragile unless done case by case.
I tested FFmpeg a lot and let me tell ya it wasn't fun dealing with the ridiculous amount of quirks.

ATRAC3/p in FFmpeg command line won't skip as it isn't implemented, in vgmstream it's done in riff.c (AT3 meta info should be "RIFF WAVE header" and not "FFmpeg supported file format").

Thealexbarney:
I've applied the padding skipping only to Switch Opus files since FFmpeg's skipping is apparently all over the place.

Based on visually inspecting the looping waveforms, the output is correct now.

bnnm:
No objections here. Probably worth investigating FFmpeg's weirdness some more, but I'm starting to think it would be simpler if I went and unified all cases in their source, though contributing involves a few too many hoops.

@kode54 - a minor crash bug was fixed in libatrac9 too, if you can update it sometime later.

Thealexbarney:
@kode54 This PR should be ready to review now

PR #213, opened 2018-04-10
Fix ATRAC9 regression

bnnm:
* Fix ATRAC9 regression 

(seems appveyor is being naughty once again, or maybe the foobar SDK changed somehow?)

ALSO! Please update libatrac9.dll, as there was a minor regression too that makes a few files decode incorrectly [>>](https://github.com/Thealexbarney/LibAtrac9/commit/91fd5507e3b1ebc88572cda68c2b7e143cb2155e)

### DIFF
--- a/src/meta/riff.c
+++ b/src/meta/riff.c
@@ -198,6 +198,7 @@ static int read_fmt(int big_endian, STREAMFILE * streamFile, off_t current_chunk
                 fmt->block_size = (bztmp & 0x3FF) * 8 + 8; //should match fmt->block_size
 #elif defined(VGM_USE_FFMPEG)
                 fmt->coding_type = coding_FFmpeg;
+                break;
 #else
                 goto fail;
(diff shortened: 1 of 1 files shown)


kode54:
I just updated libatrac9 to the latest, which is as of February something.

The SDK update broke WTL including again. Maybe you can patch that when it unpacks the dependencies?

kode54:
Ugh. He just fixed a regression today. I'll rebuild that one too.

bnnm:
Yep I noticed libatrac9 had a regression right after fixing my regression, whoops. Hopefully the last one.

Afraid I can't even get the script to work, it complains about missing .NET dependencies in my system (Win 7). Haven't really investigated since I don't build with MSVC much plus I already had my own way.
(honestly I find weird it redownloads the very latest dependencies every time, since typically one wants a build system to be repeatable with one exact working config).

PR #225, opened 2018-05-16
Add PSI file reading in Opus for loops in BlazBlue: Cross Tag Battle (Switch)

bxaimc:


src/formats.c
```diff
@@ -250,6 +250,7 @@ static const char* extension_list[] = {
     "pos",
     "ps2stm", //fake extension for .stm (to be removed)
     "psh", // fake extension for VSV(?) Dawn of Mana needs to be checked again
+    "psi", // PSI Metadata for Opus - BlazBlue: Cross Tag Battle (Switch), Maybe others?
```
kode54: Why is the extension for a metadata file being added to the open list, if it doesn't support opening that type of file directly?

bxaimc:
blargh, you're right. fixed

PR #241, opened 2018-06-22
If a file has no extension, try to play it anyway

shawwn:
This allows fb2k to play Last Of Us (PS3) speech files, which have no
file extension.

### DIFF
--- a/fb2k/foo_vgmstream.cpp
+++ b/fb2k/foo_vgmstream.cpp
@@ -341,6 +341,11 @@ bool input_vgmstream::g_is_our_path(const char * p_path,const char * p_extension
             return 1;
     }
 
+    if (strlen(p_extension) <= 0) {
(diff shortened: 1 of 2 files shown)


bnnm:
I don't know how foobar reacts to extensionless files, but for the parser I think you should use:

```
    /* check extension, case insensitive */
    if (!check_extensions(streamFile,"xvag,"))
        goto fail; 
```
meaning it'll parse "xvag" and "" (extensionless), instead of 'every single extension' which isn't very realistic.
(notice the extra comma, stuff like "ext1,,ext2" or ",ext1" are ok too)

kode54:
Waiting for this to include a proper extension check on xvag, as suggested.

shawwn:
Okay, I haven't tested it but if you're sure it'll work.

PR #246, opened 2018-07-11
vgmstream fix number 2

AnonBaiter:
adjust ps2_exst.c code because all this shit is just wrong
i just found out through Ape Escape 3's data that the EXST files do exist this way, albeit with a legit manner and with the .x extension intact
so that's why i made sure that vgmstream only plays merged EXST files with the "x" extension - i know it breaks some existing rips but it's better than nothing

by the way, the actual .sts files are stand-alone EXST headers that only serve to give in some general info to the accompanying .int file(if there is any) - due to their scattered nature however(although this must be 100% wrong but i've seen it happen most of the time in those rare PS2 games anyway) i have only seen one game(Gacha Mecha Stadium Saru Battle) in which it has both .sts and .int file in place, and only one of them at that too

here are the samples by the way(.x files only)
[https://www.sendspace.com/file/kvhb0w](url)

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -413,6 +413,7 @@ static const char* extension_list[] = {
     "wve",
     "wvs",
 
+    "x",
     "xa",
     "xa2",
     "xa30",
--- a/src/meta/ps2_exst.c
+++ b/src/meta/ps2_exst.c
@@ -24,7 +24,7 @@ VGMSTREAM * init_vgmstream_ps2_exst(STREAMFILE *streamFile) {
 
     /* check extension, case insensitive */
     streamFile->get_name(streamFile,filename,sizeof(filename));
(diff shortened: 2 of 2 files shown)


AnonBaiter:
okay, so i'll just do it like this

to ensure existing compatibility with unofficial STS files the .sts/.x file size must be checked
if said .sts/.x file size is about 0x78 then it gets a pass for being a separate .sts file - from there on vgmstream will check for an accompanying .int file, and if it exists the .int file will be used to be played with .sts info instead
otherwise said .sts/.x file will just play as normal

for an example .sts/.int file here's a sample from Gacha Mecha Stadium Saru Battle
[https://www.sendspace.com/file/o8fcwu](url)

once i know what i'm actually doing i'll be realizing this patch on full scale so just keep this pull request as-is

PR #257, opened 2018-07-17
EA support expansion

NicknineTheEagle:
* Better byte endianness auto detection.
* Properly handle dummy streams with zero offsets in BNK files.
* Added ABK header support (common format for storing SFX, contains BNK file embedded into it).

Not ready for merging yet but feel free to review and comment.


NicknineTheEagle:
So apparently, ABK files can hybrid, i.e. contain both streamed and non-streamed sounds. Shoot, I was hoping this wouldn't be the case.

NicknineTheEagle:
More findings, seems like ABK file can contain one or multiple tables (each probably representing sound category).

NicknineTheEagle:
Alright, I think I've covered everything. There's only only problem, the header point may contain duplicate entries pointing at the same tables of BNK sounds, which results in some sounds being duplicated. Not sure what to do about that.

NicknineTheEagle:
@bnnm Any idea what to do here? I guess I need to remember the tables previously parsed and discard them when they're repeated. I would use std::vector but this is pure C we're dealing with.

bnnm:
My (many) SCHl/BNK samples still work fine, all good.
I think I'll add EA SCHl ATRAC3plus next for kicks since AST are now playable.

IIRC foo_adpcm played .mus, may be of use:
https://gitlab.kode54.net/kode54/foo_adpcm/blob/master/foo_ea/mus.cpp

NicknineTheEagle:
That parser is for old NFS games with MAP/LIN/MUS files. I'm talking about the newer MPF/MUS format they've been using since around 2004.

bnnm:
> AST were playable as it was since you could use a script to cut them up into individual ASF files, just like MUS.

Erm. "since AST are now playable (directly by vgmstream, without forcing people to know the format and hunt down corresponding scripts, thus being useful enough for me to bother adding ATRAC3)".

PR #262, opened 2018-07-20
Add looping for Air (Android) Oggs and .lflac ext for kicks

bxaimc:
Add loop support for Air (Android) Oggs. Add .lflac extension for use with .pos


bxaimc:
It just looks so nice :) [https://i.imgur.com/1fEiKYT.png](https://i.imgur.com/1fEiKYT.png)

PR #265, opened 2018-07-22
More EA formats

NicknineTheEagle:
* IDX/BIG (basically a bunch of HDR/DAT compiled into one file)
* HDR/STH/DAT (used in early 7th-gen games for storing speech)
* New ABK format seen in early 7th-gen games
* Some minor tweaks

Not ready yet, ABK support incomplete, currently digging through FIFA 08 (PS3) files in an attempt to understand it further.


bnnm:
Sorry, I only had a few PSP ABKs.

NicknineTheEagle:
What about SBR/SBS, do you have any from post-2009 that hold SPS in them? Aside from FIFA 14.

bnnm:
Only that one, to save some EASPEEX samples. I've had others before (FIFA 2015 PSV, NBA Jam Wii, IIRC) but the format seemed to change around and they weren't too interesting to bother.
I think you may find some in old thread at xentax/zenhax.

NicknineTheEagle:
Ah, jeez. I guess I'll hold off on SBR/SNS for now then. Ready for merging.

bnnm:
@NicknineTheEagle - incidentally, do you have samples that would need the "streamed" flag added in ea_eaac_streamfile.h and ffmpeg_decoder_utils_ea_xma.c? Since I may unify and change stuff in the future.

NicknineTheEagle:
No, wait, bad example, these are EA-XAS. Here are some samples from FIFA 08, they're EALayer3:
https://www50.zippyshare.com/v/IyT3qg7g/file.html

Notably, the looping samples are screwed up, they cut out a few seconds early.

NicknineTheEagle:
There was an odd case where parse_variable_header only set ADPCM hist flag for newer sounds (not using codec1 tag) but then there was a special case in init_vgmstream_ea_variable_header which always assumes there's ADPCM hist for stereo EA-XA in BNKs... Anyway, this seems to be incorrect because I stumbled upon a stereo BNK sound in FIFA 98 (PC) and it has no hist so this snippet was breaking it.
And based on what I've seen so far, it looks like BNKs *never* store ADPCM hist for EA-XA. So let's run with that.

NicknineTheEagle:
Also, this should be merged into blocked layout code somehow considering ADPCM hist for most codecs is handled over there:
https://github.com/kode54/vgmstream/blob/8e16eb108cf4e85246c5036d1355fcdf0779c0ef/src/meta/ea_schl.c#L1014-L1025

NicknineTheEagle:
@bnnm - Have you seen any EA-XA BNK samples with ADCPM history? All the ones I checked don't have one.

kode54:
Let me know when this one's ready for merging, since there appear to have been commits since the last notification that it was ready.

NicknineTheEagle:
Well, it's done for now, I've just been occasionally finding more stuff to do like it always is.

bnnm:
@NicknineTheEagle - I don't think I've seen BNK with hist. The few banks I have decode the same, I guess I was being cautious.

Regarding the line [here](https://github.com/kode54/vgmstream/blob/8e16eb108cf4e85246c5036d1355fcdf0779c0ef/src/meta/ea_schl.c#L1014-L1025), I think it's better that way. I wanted to unify all the (potentially complex) logic to detect into a single variable. Otherwise I'd need to pass platform/codec/version/etc to the layout, cluttering the VGMSTREAM struct.

NicknineTheEagle:
I was thinking that you could pass ea_header struct using layout_data pointer in VGMSTREAM.

bnnm:
The layout_data is reserved for specific things so what you suggest would be quite involved to get working.

Besides, I think it makes more sense that the header parser detects that stuff once, since block layout will be called many times and that's a configuration flag that should be set before block parsing. I bet sx.exe does something similar.

PR #266, opened 2018-07-22
XA, AIF, EA-MT

bnnm:
- (dev) Handle XA ADPCM in a more standard way
- Add AIFF with XA ADPCM [Crusader: No Remorse (SAT), Road Rash (3DO)]
- Fix XA missing num_samples at the end
- Fix EA-MT full loops not being properly reset
- Add .adp AIFF extension [Sonic Jam (SAT)] 
- Fix .adp raw XA [Phantasy Star Collection (SAT)] 
- Add .ai AIFF extension [Dragon Force (SAT) sfx]

### DIFF
--- a/src/coding/coding.h
+++ b/src/coding/coding.h
@@ -82,7 +82,7 @@ size_t ps_cfg_bytes_to_samples(size_t bytes, size_t frame_size, int channels);
 void decode_hevag(VGMSTREAMCHANNEL * stream, sample * outbuf, int channelspacing, int32_t first_sample, int32_t samples_to_do);
 
 /* xa_decoder */
-void decode_xa(VGMSTREAM * stream, sample * outbuf, int channelspacing, int32_t first_sample, int32_t samples_to_do, int channel);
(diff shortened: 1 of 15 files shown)


NicknineTheEagle:
Does this fix BNK samples from SSX 3?

bnnm:
Only those that sounded bad after looping, which were about half of them I think. The rest I haven't checked yet since the EA-MT decoder is quite arcane.

AnonBaiter:
>some Crusader: No Remorse tracks have XA headers with incorrect 0xFF, rip bug/encoder feature?
hoo boy looks like you need some help
i have like 100% untouched .ACM files coming from the Saturn version of Crusader No Remorse - [here they are](https://www.sendspace.com/file/te4hm2)

bnnm:
Hmm, those look the same as samples I got from some ISO, so I guess they are just like that.
XA headers only go from 0..3 (upper nibble) and 0..C (lower nibble), so I don't know what would they mean with "0xFF". I checked Sonic Jam and they are fine, so I guess somebody messed up in Crusader.

They sound a bit harsh at points due to those glitches. Music has very few of those while voices have a lot. Oh well.

PR #268, opened 2018-07-24
vgmstream patch #3(don't mind the number skip)

RetingencyPlan:
basically, i used to have a .bms demuxer script that handled formats like "1SNh" without a hitch but now i lost it and i don't want to lurk through the hcs64 pages just to find that one script

because of this, i want to play the .UV and .TGQ files directly so forgive me if this leads to some rather disastrous errors


AnonBaiter:
whoops
turns out a simple comment-less edit caused an appveyor build failure
not sure how i managed to do that though, so... any suggestions to fix this anyone?

kode54:
I can probably force a rebuild. The stupid version reporting call needed to be changed to not care about errors like duplicate versions.

PR #275, opened 2018-08-04
Ubisoft BAO

NicknineTheEagle:
* Added support for Splinter Cell: Conviction banks for Ubi BAO parser. Incomplete - for some reason, it doesn't detect the sounds in SPK files at all. @bnnm - I could use a hand, I'm totally clueless here. Here are the samples: https://www7.zippyshare.com/v/EeESkgnD/file.html

### DIFF
--- a/src/meta/ubi_bao.c
+++ b/src/meta/ubi_bao.c
@@ -320,25 +320,26 @@ static int parse_pk_header(ubi_bao_header * bao, STREAMFILE *streamFile) {
 
     /* get stream pointed by header */
     if (bao->is_external) {
-        if (bao->prefetch_size) {
-            /* some sounds have a prefetched bit stored internally with the remaining streamed part stored externally */
-            bao->is_prefetched = 1;
+        /* some sounds have a prefetched bit stored internally with the remaining streamed part stored externally */
+        bao_offset = index_header_size + index_size;
+        for (i = 0; i < index_entries; i++) {
+            uint32_t bao_id = read_32bitLE(index_header_size + 0x08 * i + 0x00, streamFile);
+            size_t bao_size = read_32bitLE(index_header_size + 0x08 * i + 0x04, streamFile);
 
-            bao_offset = index_header_size + index_size;
-            for (i = 0; i < index_entries; i++) {
-                uint32_t bao_id = read_32bitLE(index_header_size + 0x08 * i + 0x00, streamFile);
-                size_t bao_size = read_32bitLE(index_header_size + 0x08 * i + 0x04, streamFile);
-
-                if (bao_id == bao->stream_id) {
-                    bao->prefetch_offset = bao_offset + bao->header_size;
-                    break;
-                }
-
-                bao_offset += bao_size;
+            if (bao_id == bao->stream_id) {
+                bao->prefetch_offset = bao_offset + bao->header_size;
+                break;
             }
 
-            if (!bao->prefetch_offset)
-                goto fail;
+            bao_offset += bao_size;
+        }
+
+        if (bao->prefetch_size) {
+            if (bao->prefetch_offset == 0) goto fail;
+            bao->is_prefetched = 1;
+        }
+        else {
+            if (bao->prefetch_offset != 0) goto fail;
         }
 
         /* parse resource table, LE (may be empty, or exist even with nothing in the file) */
@@ -371,13 +372,15 @@ static int parse_pk_header(ubi_bao_header * bao, STREAMFILE *streamFile) {
     }
     else {
         /* find within index */
-
         bao_offset = index_header_size + index_size;
         for (i = 0; i < index_entries; i++) {
             uint32_t bao_id = read_32bitLE(index_header_size+0x08*i+0x00, streamFile);
             size_t bao_size = read_32bitLE(index_header_size+0x08*i+0x04, streamFile);
 
             if (bao_id == bao->stream_id) {
+                /* in some cases, stream size value from 0x20 header can be bigger than */
+                /* the actual audio chunk o_O [Rayman Raving Rabbids: TV Party (Wii)] */
+                bao->stream_size = bao_size - bao->header_size;
                 bao->stream_offset = bao_offset + bao->header_size; /* relative, adjust to skip descriptor */
                 break;
             }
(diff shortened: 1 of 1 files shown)


NicknineTheEagle:
By "doesn't detect" I mean that the parser returns 414 substreams and all of them play properly but none of them are the streams from SPK file.

NicknineTheEagle:
Can confirm that it's multi-layered music tracks that are skipped.

bnnm:
Some subsongs (ex. 410) properly reference the .spk. The headers pointing to music though are being skipped though, since they use subtype 0x02 that seems some short of multilayer resource I never saw before.

NicknineTheEagle:
Yep, that's exactly what I figured.

NicknineTheEagle:
Spotted two tracks in 11_WhiteHouse_B.pk that are divided between PK and SPK. I'm guessing that these are prefetched sounds, i.e. it loads the first X ms into memory and streams the rest.

bnnm:
Note that preloading data (that I've seen) is stitched together in memory before playing (IOW may be split in the middle of a frame) so they would need [this stuff](https://github.com/kode54/vgmstream/blob/master/src/streamfile.h#L96) and isn't the easiest thing to add.

The subtype 02 seems to have stuff like tracks number, subheaders, resource size, etc, then subheaders with channels/srate/resource subsize/codec. Not fun.

NicknineTheEagle:
There are also sub-type 03 and 06, have you seen those before?

bnnm:
I think the only one I've seen is subtypes 0x04, which looked like some kind of audio config (preset?), and was also used in Ubi SB (BAO predecessor). But since subtypes didn't look very interesting and were hard to figure out I didn't pay much attention, I guess I'll add a counter+log.

NicknineTheEagle:
Currently working on implementing multi_streamfile in BAO parser.

NicknineTheEagle:
@bnnm Do you know where the size internal part for those divided files is?

bnnm:
I vaguely remember it was in some field of the header? Possibly combined/substracted with the resource size in the table. I suspect one many header flags controls if the file is split too.
Another fun feature is that XMA streams could have some kind of extra data before the actual XMA, in prefetch (.pk) and main part (.spk), one or both, possibly controled by another flag, but that's where I gave up.

If you want I can upload Scott Pilgrim PS3+X360 for cross reference, which splits all songs.

NicknineTheEagle:
Alright, this works with Splinter Cell.
@bnnm Need to update the parser for other games to get the prefetch size value.

bnnm:
Here is Scott Pilgrim [PS3](https://www.sendspace.com/file/mny33w) + [X360](https://www.sendspace.com/file/j5zi5j).

AFAIK previous versions (0x0022001E and below) don't use prefetch.

NicknineTheEagle:
Eh, which ones of these contains prefetched sounds?

NicknineTheEagle:
Does the extra data before XMA have fixed size?

NicknineTheEagle:
This works with PS3 version but not X360 because of that extra data before XMA.

bnnm:
The extra data (if you mean the stuff right before XMA data) I think was controlled by header flags, and it was a bunch of things one after another so I don't think it had a single size.

For example I've seen optional XMA header (size 0x34), various kinds tables and stuff. Possibly changed depending on the version, I think earlier games (AC, PoP, Naruto) put the XMA header before the type 0x30/0x50 data, while later games (Scott Pilgrim) put it in the type 0x20 header.

NicknineTheEagle:
Well, I've compared the same entry in PS3 and X360 files (0x00206665.pk#19) and there's one different I noticed in the sound header - there's a flag at the 0x34 which is 0 on PS3 and 1 on X360. That's likely a clue.

bnnm:
At 0x19B5C. 
A tip, find the XMA end (next 0x02250108 BAO) then go back 0x800 (XMA block size) at a time until your offset (that's assuming prefetch isn't split in the middle of a frame, which happens for AT3, but probably not for XMA). You can see XMA data usually starts like xx00xxxx too.

NicknineTheEagle:
Tried it, doesn't work. init_ffmpeg_header_offset returns NULL.

NicknineTheEagle:
Oh, extradata_offset is zero, hang on...

NicknineTheEagle:
Got it, it seems that for external sounds, XMA header is at the start of 0x30 block (the one with prefetched chunk). That still doesn't solve the issue of size.

bnnm:
[Here](https://www.sendspace.com/file/5ghfax) are some Narutos. They use XMA1 and have some funny features like some kind of multilanguage and .spk references that don't exist (maybe I'd need to rerip, or they are supposed to use the .lspk instead?)

bnnm:
I also have Assassin's Creed 1 (older and slightly different), PoP 2008 and PoP Forgotten Sands XMA, but they use the "file" BAOs which I didn't implement yet (see line ~60), since they are a pain to rip without custom .bms.

NicknineTheEagle:
Curious, language folders contain LSPK files with same names. It looks like some sort of localization functionality where it would mount different LPSK based on your language. These samples mostly work fine except for the last sound in 3c0308e1.pk which is streamed and the size of resource it uses is read incorrectly, it seems. Speech sounds are apparently referenced with a different block type than 0x20.

NicknineTheEagle:
Alright, this should do it. Figured out how to properly get XMA data start and added support for Splinter Cell: Conviciton (X360).

bnnm:
I tested it a bit and all seems in order, old and new (save the Naruto multilang, but whatevs).

NicknineTheEagle:
@kode54 I think I'm done here for now, let's get this merged.

NicknineTheEagle:
```
Downloading https://github.com/Microsoft/vswhere/releases/download/2.2.11/vswhere.exe
.\build.bat : Exception calling "DownloadFile" with "2" argument(s): "Unable to connect to the remote server"
At line:1 char:1
+ .\build.bat
+ ~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Exception calli... remote server":String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
At C:\projects\vgmstream\build.ps1:45 char:5
+     $wc.Downloadfile($uri, $outfile)
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
```
Something's up with GitHub servers, it seems.

superplayercz:
Hello, It's maybe a stupid question, but how do I convert Assassin's Creed BAO sound files?

PR #276, opened 2018-08-04
Workaround MSVC fseek/fread bug

Thealexbarney:
The MSVC runtime has a bug that results in data being returned from the wrong part of the file when using `fread` after an `fseek`.

The bug can be reproduced by decoding most HPS files with the CLI program. In the resulting output the right channel will cut out and the left channel will begin playing alternating audio from both channels.

Doing an `fseek(FILE, ftell(FILE));` works around the issue in the cases I've seen.

### DIFF
--- a/src/streamfile.c
+++ b/src/streamfile.c
@@ -76,6 +76,17 @@ static size_t read_stdio(STDIOSTREAMFILE *streamfile,uint8_t * dest, off_t offse
         else
             length_to_read = length;
 
+#ifdef _MSC_VER
+        /* Workaround a bug that appears when compiling witn MSVC.
+        * This bug is dertiministic and seemingly appears randomly
+        * after seeking.
+        * It results in fread returning data from the wrong
+        * area of the file.
+        * HPS is one format that is almost always affected
+        * by this.  */
+        fseek(streamfile->infile, ftell(streamfile->infile), SEEK_SET);
+#endif
+
         /* fill the buffer */
         length_read = fread(streamfile->buffer,sizeof(uint8_t),streamfile->buffersize,streamfile->infile);
         streamfile->validsize = length_read;


bnnm:
W.o.w. 

But wouldn't it affect like, the whole world? Maybe the trigger is something more complex? That sounds like a lot of seeking to fix it.

Also, shouldn't it be calling "fseeko" instead (like it does a bit above).

Thealexbarney:
I'm not completely sure what the trigger is. More testing or investigation is probably a good idea before applying this. I'm not sure why it pops up here and not other places. Maybe because HPS does more seeking compared to most formats?

What I do know is that there are whisperings of this sort of bug online, and that workarounds like this tend to get around it.

bnnm:
@Thealexbarney - I'll try to reproduce this and maybe compare with GCC, do you have some HPS that always triggers it? Doesn't look like HPS is doing more seeking that any other blocked files tho. Maybe related to decoding too fast/SSD drive/Windows version? Also foobar is doing its own read/seek, maybe could try to compare its wav writer.

@kode54 - I think they are buffered ok but it would always happen as it's reading/decoding the whole file I guess. Some blocked layouts or codecs can trash the buffers a lot, I'll see how many reads are being done.

Thealexbarney:
@bnnm I don't know that I've seen any HPS files that don't trigger it. `1p_qk.hps` from SSBM triggers at ~5 seconds.

bnnm:
Ok, I compiled test.exe with VC2015 (not sure which sub-version) and it outputs the same exact file that GCC does. I take appveyor is using VC2018? Maybe some update could fix the bug?

WTF Microsoft, first [this](https://yro.slashdot.org/story/16/06/10/1350245/visual-studio-2015-c-compiler-secretly-inserts-telemetry-code-into-binaries), now basic IO crap.

kode54:
foobar2000 also does not use stdio, it uses its own IO replacement, so it can use its IO services for everything, which internally use Win32 IO directly.

bnnm:
Did some searching but not much info around ([link](https://stackoverflow.com/questions/43732499/vs2015-non-accurate-fread-after-a-fseek-operation), [link](https://social.msdn.microsoft.com/Forums/Lync/en-US/8c4d27db-1128-4803-951b-fef515455a03/vs2015-non-accurate-fseekfread-operation?forum=vcgeneral)).

Anyway, if it fixes MSVC's test/winamp builds I think it should be accepted. Reads are buffered so any performance penalty will be minimal.

It saddens me a bit the fix is black magic tho. Maybe the ifdef could be restricted later to _MSC_VER known to fail, as my MSVC builds work fine.

PR #279, opened 2018-08-08
EA formats

NicknineTheEagle:
Some organization and improved EA ABK to support more games. Maybe more stuff to come.

### DIFF
--- a/src/layout/blocked_ea_schl.c
+++ b/src/layout/blocked_ea_schl.c
@@ -23,7 +23,7 @@ void block_update_ea_schl(off_t block_offset, VGMSTREAM * vgmstream) {
     {
         uint32_t block_id = read_32bitBE(block_offset+0x00,streamFile);
 
-        if (guess_endianness32bit(block_offset + 0x04,streamFile)) /* size is always LE, except in early SS/MAC */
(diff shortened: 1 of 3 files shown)


NicknineTheEagle:
Welp, turns out that that Wii/3DS games use HDR/DAT with *new* HDR header but *old* audio format. This is just messy, it makes HDR parsers unreliable. I'm really not sure what to do about this, both HDR headers are so simple you can't really check anything on them.

bnnm:
Maybe you could test the HDR pointed data then call old/new stuff?

In case it helps to give you ideas, [here](https://github.com/kode54/vgmstream/blob/master/src/meta/sps_n1.c) is an example of calling parsers withing parsers using custom IO. 
It makes metas cleaner (not having to pass base offset/etc, and you can alter the resulting vgmstream), though could be harder to follow.

NicknineTheEagle:
I was thinking about that and it won't do. I found these new HDR headers in FIFA 12 (Wii) and Wii games use the same SCHl audio format as 6th gen and it also re-uses GC platform ID. Audio codec is also same (EA MicroTalk). So the pointed audio streams are indistinguishable from the ones you'll see on Gamecube.

NicknineTheEagle:
@kode54 No more upcoming changes for now, merge please.

PR #280, opened 2018-08-08
vgmstream fix number 4

AnonBaiter:
this pull request adds extension-less support to the SShd/SSbd format as seen in MotoGP by Namco

samples to test are [here](https://www.sendspace.com/file/lanvtp)


AnonBaiter:
to compensate for the wait i fixed a rather nasty bug regarding VPK files
play these files on vgmstream's current version and you'll see why

[https://www.sendspace.com/file/is1ykc]([NBA08_PS3_demo]4ch_VPK.7z)

PR #302, opened 2018-09-30
g7221

bnnm:
- Update G.722.1C (Siren14) decoder DLL to fix volume and others
  - Note: must re-add .asc file

***

I wasn't satisfied with the current libsiren DLL output so I did a wrapper for the float ref.decoder (see here for the code: https://github.com/bnnm/vgmstream-g7221).

Main libsiren problems that this DLL fixes:
- Siren14 volume was quieter vs reference int decoder
- output could change slightly when re-decoding the same file (threading?)
- had some bugs and edge cases fixed in later versions of the ref.code
- even though it mostly repurposed the ~2000 v1 ref.code, libsiren removed all Polycom credit and comments (not cool)

The DLL uses the float ref.code (like libsiren), performance is very similar (slightly faster, perhaps).
I did try to optimize int decoder, got around ~25% faster but it was still about 4 times slower.

I decided to leave the ref.code almost untouched (after some fixes), ugly as it is, since I'd rather make a new decoder from scratch for vgmstream than having some trimmed franken-code (may be necessary later, as some games use an updated version of Siren14, could remove encoder/Siren7 part and fix BE reading). Shouldn't affect compiled DLL anyway.

Ref.decoder didn't do float Siren14 originally but I patched it along thread safeness and stuff (see commit history).


bnnm:
To clarify, the ref.decoder (Siren7+14) comes in two flavors, "int" and "float" (as in, one uses int math and other float math):
- The "int" is accurate, but super slow (it does a ton of extra operations that I can't figure out how to optimize). It was the one hcs used long ago.
- The "float" is fast (same libsiren used), slightly less accurate. Code needed some fixes first though.

I did a lot of tests of hcs's int vs libsiren's float vs this DLL's float, it works well I promise.

PR #306, opened 2018-10-09
Add Konami / Sony ATRAC3 support for VA3 files

chrispable:
I would probably REJECT commit 107f9a6 and just add the file to the project yourself, my version of visual studio is weird and added awkward stuff to the file so I kept it separate. Plus then you dont need to change the platform toolset. This should be an easy right-click -> include in project :).

### DIFF
--- a/cli/vgmstream_cli.c
+++ b/cli/vgmstream_cli.c
@@ -542,8 +542,13 @@ int main(int argc, char ** argv) {
     return EXIT_SUCCESS;
 
 fail:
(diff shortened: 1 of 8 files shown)


kode54:
For future reference, please hand edit your project files before committing them to our tree, to edit out any such changes.

chrispable:
Sorry about that. I actually didn't even realize the project file was plain text until I saw a diff. My bad entirely.

PR #334, opened 2018-12-21
Ubi SB: SMx format support

NicknineTheEagle:
I have written a script for MAPS.SMx format that cuts them up into fake SBx files: https://pastebin.com/vpmmACE1 With some tweaks to the parser, they're playable in vgmstream and I have added support for section 2 entries and codecs used in some games with that format.

@bnnm I'd like to hear your thoughts on this idea. This may be at least good as a temporary solution for testing those games.


NicknineTheEagle:
> Splitting files is passable I guess but I definitely dislike very much data conversions. If you add that for testing, I ask you to at least mark which games and codecs are map-only and document "is_map" is for bms-created fake files that don't actually exist in the wild.

Yeah, I can add some doc to the code.

> As for another alternative for easier usage, I'd rather have a bms that splits a map with multiple sections into a map with a single section, which should be a lot cleaner and simpler to do. Then common SBx and SMx parts could be reused/refactored instead, in the same file even.

Ah, that's a good idea, I'll see what I can do.

NicknineTheEagle:
Although in that case, we might as well add full support for SMx format since single map files split up from a big one aren't going to be any different in format.

bnnm:
Sure, though the main issue about full SMx that is that you'd need to count all sounds from all sections to get total subsongs, and would be in the thousands and wouldn't be very usable. But having full support and a splitter bms for easier usage would be best of both worlds I think.

NicknineTheEagle:
Here we go, I've added full support for SMx format. I've also updated the script so it splits SMx files into single map SMx files: https://pastebin.com/vpmmACE1

PR #339, opened 2018-12-27
Ea map mpf

NicknineTheEagle:
This should be pretty much every relevant EA bank/playlist format. As such, I've removed the hack in SCHl parser where bigfiles with multiple sounds are stitched into one track, it's rendered entirely obsolete by full soundbank support and TXTP.
Also removed IDX/BIG since it's not really a bank but a bunch of HDR/DAT banks compiled into a single file, you get thousands of subsongs if you load it as is. I've made it a parser for it here: https://pastebin.com/VXFRxFQq


NicknineTheEagle:
Gotta fix some SSX 3 files.

NicknineTheEagle:
This is very suboptimal solution, not only it's phased out by TXTP but the sounds in MUS files are rarely in a convenient order. It should really go, IMO.

bnnm:
When I update a program I *hate* when things stop working for no apparent reason. Thus I'm very, very careful not to break things unnecessarily.

So if I have a bunch of EA sets around, after your update they'll stop working. Now I'll have to:
- re-download a bunch of potentially gigantic games, some that may be hard to find working links
- install them all, then possibly using who-knows-what tools and bms to rip
- reupload all (which is seemingly impossible now)
- ask everybody to redownload

...why? Don't you think this is a massive dick move?

Can't you just have MPF+TXTP for newer, good rips, and that hack for older, crap rips?

NicknineTheEagle:
Sorry, I was not aware that so many that so many sets relied on it, if it's really that big of an issue, I can restore the hack.
Just how many sets use it, though? I'm looking through NFS and SSX 3 sets and I'm seeing sets of fake EAM files, is that how a lot of them are set up?

bnnm:
Thanks, as for sets IIRC a bunch of Harry Potters, GoldenEye Rogue Agent, and early NFS needed it.

Just to make it clear I'm not against removing unused/crap code, but only when I'm reasonably sure it isn't needed anymore. I've broken stuff in the past but I've provided replacements rips first when possible and made it clear to give people a chance to redownload.

Also I suggest you use subfile streamfiles for those bank formats [(example),](https://github.com/kode54/vgmstream/blob/0bc8e084cf4a1b7f6cde3bbc68b35fc7b146835f/src/meta/imc.c#L69) as it makes metas a lot cleaner (plus I think it mimics sound engine internals better).

NicknineTheEagle:
> Also I suggest you use subfile streamfiles for those bank formats (example), as it makes metas a lot cleaner (plus I think it mimics sound engine internals better).
There is a problem with this: calculating size for BNK sounds is not possible, there are no sizes in the header, no block sizes, nothing.

kode54:
Is this ready?

NicknineTheEagle:
Yeah, I'm done.

PR #370, opened 2019-03-17
Add CMake scripts

CyberBotX:
Needed to change how version.h was included to let it pick up the one from CMake's build location.

NOTE: Does not build with MinGW-w64 currently, does not have a way to build the audacious or XMPlay plugins.


CyberBotX:
Oh, something I forgot to mention in the initial pull request, and probably something I should add to the root CMakeLists.txt, but because of the use of file(GLOB), whenever any files are added or removed from libvgmstream, CMake will need to be re-run in order to pick up the file changes, it will not pick them up otherwise.

CyberBotX:
Unless anyone has changes or suggestions to all of this, I think this is ready for merging. Hopefully CMake should be helpful to others as much as it has been for me.

awsms:
Just needed to ask, is Audacious 3.5 still the last version supported by vgmstream ?

CyberBotX:
From what I've heard, I believe that is the case. bnnm had said that on Discord and that he tested with 3.7+, and I tested with 3.10.

awsms:
@CyberBotX Thanks for the answer. I tried with 3.10 as well yesterday, build seems ok but the plugin doesn't appear on Audacious. Will have to inspect it closely

PR #392, opened 2019-04-09
fixes

bnnm:
- Fix .hca key detection for newer Tales of the Rays files
- Fix some .msf/mp3 looping


bnnm:
Merged this (a bit as a test), since they were simple fixes

PR #421, opened 2019-06-12
Add 3DS Ubi RAKI .ckd [Gravity Falls (3DS)]

bnnm:


Danominator:
Hi! I'm the guy who requested this file extension to be added for the Gravity Falls game. Thanks for the help with that by the way!

I have noticed that the quality of the sounds tends to be lowered greatly when exporting to mp3. Is there another format I should be using in order to get the quality closest to what foobar provides?

bnnm:
WAV or FLAC would be best, or maybe high quality settings with lame.exe MP3, or OGG/OPUS should be ok too (you need to find and install appropriate .exes though).

Danominator:
I used this, and with high quality settings, they still came out pretty bad as mp3. I feel like mp3 should be more clear than this.

https://www.foobar2000.org/encoderpack

You can try it yourself on this file:

[music_forest_dark.wav.zip](https://github.com/losnoco/vgmstream/files/3497810/music_forest_dark.wav.zip)

Danominator:
WAV and FLAC still cause a loss in quality it seems. I must be doing something wrong in the settings.

kode54:
Sounds fine to me, but what great things are you expecting from a 24 kHz source file?

Danominator:
Well, that's the weird thing. They sound great in foobar before exporting. Then it comes out more muffled when exported. The sound effects suffer from this especially, like this one.

[sfx_jingle_newtoolweapon.wav.zip](https://github.com/losnoco/vgmstream/files/3499628/sfx_jingle_newtoolweapon.wav.zip)

kode54:
That one's 16 kHz. At this point, I'd try my luck with upsampling them to 44100 Hz, which, while technically upsampling, should not add any extra information.

With that first sample you posted, it increases the bitrate of the resulting file by 20kbps, but doesn't result in a noticeably different file.

The jingle sounded about the same to my ears, but my hearing isn't that great any more.

PR #460, opened 2019-08-17
fix opus sample rate for Grandia HD Collection (Switch)

bxaimc:
Header contains inaccurate sample rate of 16000 Hz, needs to be 48000 Hz.


AnonBaiter:
or you could just force vgmstream to read all opus files as 48000 if we're talking about sample rate here, considering opus encodes everything to 48000 sample rate frequency no matter what.

PR #473, opened 2019-09-15
faketxtp

bnnm:
- Allow virtual .txtp in playlists (foobar/winamp/cli) for config/tagging

@kode54 - if you can take a look at the foobar code in case I did something dumb, though it works ok (if unorthodox).

### DIFF
--- a/fb2k/foo_streamfile.cpp
+++ b/fb2k/foo_streamfile.cpp
@@ -20,6 +20,7 @@ extern "C" {
 typedef struct {
     STREAMFILE sf;          /* callbacks */
 
+    bool m_file_opened;         /* if foobar IO service opened the file */
     service_ptr_t<file> m_file; /* foobar IO service */
     abort_callback * p_abort;   /* foobar error stuff */
     char * name;                /* IO filename */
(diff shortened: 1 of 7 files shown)


fb2k/foo_streamfile.cpp
```diff
@@ -193,16 +199,23 @@ static STREAMFILE * open_foo_streamfile_buffer_by_file(service_ptr_t<file> m_fil
...
+    if (infile_exists) {
+        filesystem::g_open_read(infile,filename,*p_abort);
+        if(stats) *stats = infile->get_stats(*p_abort);
+    }
 
-    streamFile = open_foo_streamfile_buffer_by_file(infile,filename,buffersize,p_abort);
+    streamFile = open_foo_streamfile_buffer_by_file(infile, infile_exists, filename, buffersize, p_abort);
     if (!streamFile) {
-        // fclose(infile);
+        //m_file.release(); //todo not needed after g_open_read?
```
kode54: service_ptr_t is reference counting and auto releasing when it goes out of scope. Manual releasing is only necessary when it's a member of a structure that wasn't allocated with new. Gotta make sure it's initialized in place in that case.

PR #500, opened 2019-11-04
cleanup, afc, isb

bnnm:
- Improve raw MP3 total samples reading
- Add sfx/str/zic CWAV extensions
- Move some extensions to normal list
- Clean some winamp code
- Improve AFC/DTK/FADPCM/PROCYON/LSF/ASF/DSF/XMD/MTAF/MTA2 performance and cleanup
- Remove stx parser since it's a repeat of afc
- Improve streamfile float reading
- Add extra MPF+MUS pair and missing EA .hdr check
- Fix uncommon handle leak when opening too many files
- Tweak _0/1 dual extension detection
- Add ISACT .isb [Psychonauts (PC)]
- Allow coef_offset to work with subsongs in TXTH
- Add HCA key [Blackstar (Mobile)]
- Fix FFmpeg slowness in MSVC
- Improve SCHl MP2/EALayer3 V1a gapless


bnnm:
@kode54 I'm not sure what was changed but I'm getting lots of line breaks diffs I never had before

bnnm:
Ok, you can see diffs without line breaks by adding `?w=1` to the url.

I'll just merge and in the future maybe commit LF changes first then normal changes per file, since changing to LF the whole thing may polute history a bit too much.

PR #544, opened 2020-01-07
Added FWSE format.

PredatorCZ:
Can be now used as `add_subdirectory` for CMake.

### DIFF
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -17,8 +17,8 @@ if(WIN32)
 
 	# Make sure that the binary directory is included (for version.h), as well as the getopt library include directory
 	target_include_directories(vgmstream_cli PRIVATE
-		${CMAKE_BINARY_DIR}
-		${CMAKE_SOURCE_DIR}/ext_libs/Getopt)
+		${VGM_SOURCE_DIR}
+		${VGM_SOURCE_DIR}/ext_libs/Getopt)
 
 	# Include the version string
 	if(MSVC)
@@ -44,7 +44,7 @@ endif()
 
 # Install the CLI program
 install(TARGETS vgmstream_cli
-	RUNTIME DESTINATION bin)
+	RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
 
 # TODO: Make it so vgmstream123 can build with Windows (this probably needs a libao.dll included with vgmstream, though)
 
@@ -70,5 +70,5 @@ if(NOT WIN32)
 
 	# Install vgmstream123
 	install(TARGETS vgmstream123
-		RUNTIME DESTINATION bin)
+		RUNTIME DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
 endif()
--- a/cli/CMakeLists.txt
+++ b/cli/CMakeLists.txt
@@ -17,7 +17,7 @@ if(WIN32)
 
 	# Make sure that the binary directory is included (for version.h), as well as the getopt library include directory
 	target_include_directories(vgmstream_cli PRIVATE
-		${VGM_SOURCE_DIR}
+		${VGM_BINARY_DIR}
 		${VGM_SOURCE_DIR}/ext_libs/Getopt)
 
 	# Include the version string
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -5,10 +5,10 @@ file(GLOB LAYOUT_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/layout/*.c")
 file(GLOB META_HEADERS "${CMAKE_CURRENT_SOURCE_DIR}/meta/*.h")
 file(GLOB META_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/meta/*.c")
 set(EXT_HEADERS
-	${CMAKE_SOURCE_DIR}/ext_includes/clHCA.h
-	${CMAKE_SOURCE_DIR}/ext_includes/pstdint.h)
+	${CMAKE_CURRENT_SOURCE_DIR}/../ext_includes/clHCA.h
+	${CMAKE_CURRENT_SOURCE_DIR}/../ext_includes/pstdint.h)
 set(EXT_SOURCES
-	${CMAKE_SOURCE_DIR}/ext_libs/clHCA.c)
+	${CMAKE_CURRENT_SOURCE_DIR}/../ext_libs/clHCA.c)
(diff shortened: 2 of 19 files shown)


CyberBotX:
Other than the change in the root CMakeLists.txt, I'm not sure I see why the other CMake changes are necessary? The existing paths should be working fine, I've seen no issues with them as they are.

bnnm:
Code looks fine (can't comment on CMake stuff) but if you could try to merge first, I'm no good at that.
Also maybe rename `cpm_ima_expand_nibble` to `mft_ima_expand_nibble` for consistency.

PredatorCZ:
> 
> 
> Other than the change in the root CMakeLists.txt, I'm not sure I see why the other CMake changes are necessary? The existing paths should be working fine, I've seen no issues with them as they are.

They use CMAKE_SOURCE_DIR, this won't work, if they're used as subdirectory. Since this variable will be always location of very root CMakeLists.txt.
I can modify `set(VGMSTREAM_EXT_INCLUDES ${CMAKE_CURRENT_SOURCE_DIR}../ext_includes)` for better clarification or replace this var directly with `${CMAKE_CURRENT_SOURCE_DIR}../ext_includes/...`, whichever you prefer better.

CyberBotX:
> > Other than the change in the root CMakeLists.txt, I'm not sure I see why the other CMake changes are necessary? The existing paths should be working fine, I've seen no issues with them as they are.
> 
> They use CMAKE_SOURCE_DIR, this won't work, if they're used as subdirectory. Since this variable will be always location of very root CMakeLists.txt.
> I can modify `set(VGMSTREAM_EXT_INCLUDES ${CMAKE_CURRENT_SOURCE_DIR}../ext_includes)` for better clarification or replace this var directly with `${CMAKE_CURRENT_SOURCE_DIR}../ext_includes/...`, whichever you prefer better.

I understand that CMAKE_SOURCE_DIR is the directory of the root CMakeLists.txt, but I was referencing ext_includes relative to that directory, so that's why I'm confused as to how this won't work. It's been working fine for me from what I've seen. In what cases does it not work out?

PredatorCZ:
> I understand that CMAKE_SOURCE_DIR is the directory of the root CMakeLists.txt, but I was referencing ext_includes relative to that directory, so that's why I'm confused as to how this won't work. It's been working fine for me from what I've seen. In what cases does it not work out?

When I use it as `add_subdirectory(modules/vgmstream)`.

CyberBotX:
> > I understand that CMAKE_SOURCE_DIR is the directory of the root CMakeLists.txt, but I was referencing ext_includes relative to that directory, so that's why I'm confused as to how this won't work. It's been working fine for me from what I've seen. In what cases does it not work out?
> 
> When I use it as `add_subdirectory(modules/vgmstream)`.
> Or I'm doing something wrong.

Where are you doing the add_subdirectory from and what problems are being caused by the current CMake code? I might be able to help figure out why but not without knowing what the problem actually is, because it doesn't make sense to me currently.

PredatorCZ:
Please check https://github.com/PredatorCZ/RevilToolset/blob/0f8a1bb4061d43775e6cd3cca16f94b462e7880a/SPACConvert/CMakeLists.txt#L35

This is how I use it.

CyberBotX:
Ah, I see now. OK, then yes, please make sure all paths use `CMAKE_CURRENT_SOURCE_DIR` to keep things relative. Not using that could lead to some subtle errors I believe. If you are going to do this, please do it throughout the project. Also, I believe you `CMAKE_CURRENT_SOURCE_DIR` does not include a trailing slash so be sure to put a slash after it.

PredatorCZ:
Added more relativity.
Tested with WIN32 and install command.
Following are untested:
Maiattrac3+
Foobar plugins (uses vcxproj)
Audacitous

Also a question, where do you want to generate mp4v2 config header?
I believe that you use CMAKE_SOURCE_DIR == CMAKE_BINARY_DIR, not everyone uses such approach, so you want to keep it in source dir, or move it to the binary dir?

CyberBotX:
The reason that I made the mp4v2 header go to the binary dir is because I focused on out-of-source builds to keep the source tree pristine. Writing to the source directory is a bad approach in my opinion, as it makes source control harder to deal with.

PredatorCZ:
Yes, I totally agree. I just got a bit confused by CI builds and the fact, that you generate version files in the source dir. Changed to binary dir.

cli/CMakeLists.txt
```diff
@@ -17,8 +17,8 @@ if(WIN32)
 
 	# Make sure that the binary directory is included (for version.h), as well as the getopt library include directory
 	target_include_directories(vgmstream_cli PRIVATE
-		${CMAKE_BINARY_DIR}
-		${CMAKE_SOURCE_DIR}/ext_libs/Getopt)
+		${VGM_SOURCE_DIR}
```
CyberBotX:
```suggestion
		${VGM_BINARY_DIR}
```

src/CMakeLists.txt
```diff
@@ -41,7 +41,8 @@ setup_target(libvgmstream)
 
 # Set up the proper include directories
 target_include_directories(libvgmstream PRIVATE
-	${CMAKE_CURRENT_SOURCE_DIR}
+	${VGM_SOURCE_DIR}
```
CyberBotX:
```suggestion
	${VGM_BINARY_DIR}
```

winamp/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ target_compile_definitions(in_vgmstream PRIVATE
 	_UNICODE)
 
 # Make sure that the binary directory is included (for version.h)
-target_include_directories(in_vgmstream PRIVATE ${CMAKE_BINARY_DIR})
+target_include_directories(in_vgmstream PRIVATE ${VGM_SOURCE_DIR})
```
CyberBotX:
```suggestion
target_include_directories(in_vgmstream PRIVATE ${VGM_BINARY_DIR})
```

src/CMakeLists.txt
PredatorCZ: Doesn't seem that this lib requires such include.
I can build it without this path, please confirm.

CyberBotX: If it build fine without it, then it is fine to leave it out. I could've sworn there was a reason I had it set... (and it should've been SOURCE, not BINARY, my suggested change is wrong in this case).

PredatorCZ: It's already removed.
I think this MR can be resolved.

PR #568, opened 2020-03-02
Better decode for PSX & PSV ADPCM

korenkonder:
* Separated float and int math for extended mode in PSX decoder
* Clamping only output samples

### DIFF
--- a/src/coding/psv_decoder.c
+++ b/src/coding/psv_decoder.c
@@ -173,25 +173,25 @@ void decode_hevag(VGMSTREAMCHANNEL * stream, sample_t * outbuf, int channelspaci
     if (shift_factor > 12)
         shift_factor = 9; /* ? */
 
+    shift_factor = 20 - shift_factor;
     /* decode nibbles */
     for (i = first_sample; i < first_sample + samples_to_do; i++) {
-        int32_t sample = 0, scale = 0;
+        int32_t sample = 0;
 
         if (flag < 0x07) { /* with flag 0x07 decoded sample must be 0 */
             uint8_t nibbles = frame[0x02 + i/2];
 
-            scale = i&1 ? /* low nibble first */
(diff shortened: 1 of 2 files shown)


bnnm:
Could you provide rationale/references for your changes?

Implementations and variations of this codec sometimes do need hist clamping to get correct output. For example Namco's implementation (PC, probably PS3) doesn't clamp it, while EA's (some PC implementations) does clamp it.

The way shift + sign extend and other code was handled in PSX is very common in these codecs and I purposedly left as documentation, even though it can be simplified. Perhaps that and such code were changed for micro-optimizations, but from personal tests this has little to no effect in performance (consider stuff like compiler optimizations, CPU cache and branch prediction).

Also float_mode was left for future changes. Whether the codec uses float or int modes is not related to the extended_mode but platform/implementation (PS3/later PC games normally use non-extended mode and float in sw).

korenkonder:
Okay. I can show you how differs output from my edit.

vgmstream_cli was built with latest Windows 10 SDK and v141 toolset

My build:
image
Current vgmstream build:
image

Did you noticed that strange "peaks" in current realization's output diff?

Making difference between original and compressed audio
image

Archive with build and test files:
https://yadi.sk/d/UM9S6_9XKKBqzw

AnonBaiter:
as far as "PS1/PSV decoding algorithms" go, this pull request seems good enough for me.
with that being said, i care more about how "faithful" or (even better) "accurate" it (the recently-changed decoding algorithm) is.

korenkonder:
What codecs uses this ADPCM technique? I can do that stuff for all of them

bnnm:
Thanks, the removal of clamping makes sense. I tested with regular PS-ADPCM for a bunch of games and only few files seem to change output though.

The extended part I'd prefer if you could revert it, since I plan to change stuff around in the future, so I'd have to do it anyway.

PR #571, opened 2020-03-03
Vag enhance

korenkonder:
* Separated float and int math for extended mode in PSX decoder
* Clamping only output samples

[Old pull request](https://github.com/losnoco/vgmstream/pull/568)

> vgmstream_cli was built with latest Windows 10 SDK and v141 toolset
> 
> My build:
> image
> Current vgmstream build:
> image
> 
> Did you noticed that strange "peaks" in current realization's output diff?
> 
> Making difference between original and compressed audio
> image
> 
> Archive with build and test files:
> https://yadi.sk/d/UM9S6_9XKKBqzw
> 
> P.s. If you want, I can revert changes for extended part
> P.p.s. Audio was compressed by currently-in-dev VAGAudio encoder, which is "private" now

### DIFF
--- a/src/coding/psv_decoder.c
+++ b/src/coding/psv_decoder.c
@@ -173,25 +173,25 @@ void decode_hevag(VGMSTREAMCHANNEL * stream, sample_t * outbuf, int channelspaci
     if (shift_factor > 12)
         shift_factor = 9; /* ? */
 
+    shift_factor = 20 - shift_factor;
     /* decode nibbles */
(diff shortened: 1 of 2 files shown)


korenkonder:
Maybe this SPU-ADPCM patch can be applied to other formats like FADPCM

bnnm:
AFAIK this does not apply to FADPCM, that must clamp hist and gives same results vs FMOD PC tools. Note that some codecs that are marked as reverse engineered, as in, code should be correct already.

PR #636, opened 2020-05-28
Add support for OGG LoopFlags when using ffmpeg and fix a link issue

julianxhokaxhiu:
I've being using the FFMpeg patch for a while now in my project ( https://github.com/julianxhokaxhiu/FFNx ) and it works extremely well with Ogg files that do have those metadata. I do believe it will work with any file that can contain metadata which FFMpeg can access.

I'd be happy if this could be merged upstream so I don't need to maintain a separate fork of VGMStream and I can continue to build it upstream :)

The second patch is a minor link issue I detected that will not allow my project to link successfully to vgmstream because `init_vgmstream_mups` requires `init_vgmstream_ogg_vorbis` to be available. When you do not define `USE_VORBIS` while building though, it will fail. I think placing the call behind the ifdef will provide a stable solution. Feel free to let me know if another road has to be chosen.

Thank you!

### DIFF
--- a/src/meta/ffmpeg.c
+++ b/src/meta/ffmpeg.c
@@ -17,7 +17,7 @@ VGMSTREAM * init_vgmstream_ffmpeg_offset(STREAMFILE *streamFile, uint64_t start,
     VGMSTREAM *vgmstream = NULL;
     ffmpeg_codec_data *data = NULL;
     int loop_flag = 0;
-    int32_t loop_start = 0, loop_end = 0, num_samples = 0;
+    int32_t loop_start = -1, loop_end = -1, num_samples = 0;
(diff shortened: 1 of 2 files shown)


NicknineTheEagle:
This code is never actually reached unless you compiled vgmstream without libvorbis since there's already OGG parser which handles loop flags: https://github.com/losnoco/vgmstream/blob/master/src/meta/ogg_vorbis.c

NicknineTheEagle:
I'm not sure if it's productive to duplicate all the loop point handling in OGG and FFmpeg parsers.

julianxhokaxhiu:
That's although my personal issue. Maybe we can find a better way to abstract those checks across all codecs. But at this stage as I'm using only FFMpeg as backend for VGMStream, works fine for me.

Issue #674, opened 2020-07-24
Some Wii .brstm files have the right channel incorrectly decoded


chenxiaolong:
It seems that with some `.brstm` files, vgmstream isn't decoding the right channel properly. For example:

* Original `.brstm`: https://1drv.ms/u/s!An4x47eHOD-mkF-AGeLAkqiW4-Tk?e=5YcEjP
* `.wav` converted by brawlbox: https://1drv.ms/u/s!An4x47eHOD-mkGCd_4vQJrPwSBc8?e=DM30Lh
* `.wav` converted by vgmstream: https://1drv.ms/u/s!An4x47eHOD-mkF4ZlK8BBRcf3xsZ?e=l0y1dh

This is what the two waveforms look like:

![image](https://user-images.githubusercontent.com/646253/88439727-02195280-cdda-11ea-9e29-635f52e7ce32.png)

Briefly comparing the two `.wav` files with vbindiff seems to confirm that only the right channel is affected since only every other 2 bytes is different. I'm not immediately seeing what is wrong. It doesn't seem to be a simple bit flip or bit shift issue.

Any ideas what might be going on? I was testing with commit 3db9b74e4c278cd9e8ee616a426998071672cf92 on Linux (gcc 9.3.0) and VS2019.

(By the way, I don't think this is a newly introduced bug. I tried building a very old commit from 2012 in an attempt to git bisect this, but it seems it was always an issue.)

![image](https://user-images.githubusercontent.com/646253/88439777-29701f80-cdda-11ea-9fa3-a42bce19a242.png)

chenxiaolong:
I think it might be due to something here: https://github.com/losnoco/vgmstream/blob/3db9b74e4c278cd9e8ee616a426998071672cf92/src/meta/brstm.c#L126-L130

If I break out of the outer loop after the 1st iteration, I get audio on the right channel that somewhat resembles what it should be.

chenxiaolong:
These are the coefficients produced by the two programs. I'm still working on figuring out where the difference in calculation is.

| Coeff index | vgmstream | brawlbox |
|-------------|-----------|----------|
| `[0][0]`    | `0x02ba`  | `0x02ba` |
| `[0][1]`    | `0x003e`  | `0x003e` |
| `[0][2]`    | `0x06a1`  | `0x06a1` |
| `[0][3]`    | `0xfe80`  | `0xfe80` |
| `[0][4]`    | `0x0020`  | `0x0020` |
[...]

PR #675, opened 2020-07-25
meta/brstm.c: Parse ADPCM header offset for each channel instead of skipping a fixed distance

chenxiaolong:
The parser previously computed the offset of the coefficients field of
channel 0's ADPCM header and then skipped 0x38 bytes to get to the
coefficients field of the next channel's ADPCM header. This breaks for
some files because they have larger 0x3a byte ACPCM headers.

This commit updates the logic to compute the offset of each channel's
ADPCM header. Immediately after the HEAD part 3 header, there's an
8-byte table for each channel, where the second field indicates the
offset of that channel's ADPCM header. With this information, it's just
a matter of adding 0x08 to get the coefficients field.

Fixes: #674

Signed-off-by: Andrew Gunnerson <chillermillerlong@hotmail.com>

### DIFF
--- a/src/meta/brstm.c
+++ b/src/meta/brstm.c
@@ -106,26 +106,39 @@ VGMSTREAM * init_vgmstream_brstm(STREAMFILE *streamFile) {
 
     if (vgmstream->coding_type == coding_NGC_DSP) {
         off_t coef_offset;
-        off_t coef_offset1;
-        off_t coef_offset2;
+        off_t head_part3_offset;
+        off_t adpcm_header_offset;
         int i,j;
-        int coef_spacing = 0x38;
+        int coef_spacing;
 
         if (atlus_shrunken_head)
         {
             coef_offset = 0x50;
             coef_spacing = 0x30;
+
+            for (j = 0; j < vgmstream->channels; j++) {
+                for (i = 0; i < 16; i++) {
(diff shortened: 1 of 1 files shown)


chenxiaolong:
I intentionally kept the old logic in place if `atlus_shrunken_head` is true because I don't know if those files have the tables indicating where each ADPCM header is. If someone could send me a sample file, I could check it and update the code accordingly.

chenxiaolong:
I was able to get a copy of Trauma Center: Second Opinion and it looks like those brstm files are missing the fields that point to each channel's ADPCM headers. The old logic will remain in place if `atlus_shrunken_head` is true.

---

I have now tested this PR with a few games:
* Trauma Center: Second Opinion - both the `.brstm` files with broken headers and the ones with good/normal headers
* New Super Mario Bros Wii - every `.brstm` file
* Newer Super Mario Brows Wii ROM hack - every `.brstm` file and every `.er` file (renamed `.brstm`)
* Super Smash Bros Brawl - every `.brstm` file

bnnm:
Note that the same logic applies to [bfstm](https://github.com/losnoco/vgmstream/blob/master/src/meta/bfstm.c#L115) and similar Nintendo headers, but some still need to be cleaned up.

PR #688, opened 2020-08-16
Highly improved Game.com ADPCM decoding algorithm

simontime:
I improved and optimised the decoding algorithm, it also incidentally fixes the DC offset issue.


simontime:
Don't merge just yet, I found 1 file there's a small decoding hiccup with

PR #701, opened 2020-09-06
txtp mul

bnnm:
- Fix some loop install issues with layers
- Fix some .mul [Tomb Raider 8 (PS3), Avengers (PC)]

### DIFF
--- a/src/meta/mul.c
+++ b/src/meta/mul.c
@@ -3,15 +3,15 @@
 #include "../coding/coding.h"
 #include "mul_streamfile.h"
 
-typedef enum { PSX, DSP, IMA, XMA1 } mul_codec;
+typedef enum { PSX, DSP, IMA, XMA1, FSB4 } mul_codec;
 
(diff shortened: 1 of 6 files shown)


src/meta/mul.c
```diff
@@ -48,21 +48,23 @@ VGMSTREAM * init_vgmstream_mul(STREAMFILE *sf) {
     /* 0x34: id? */
     /* 0x38+: channel config until ~0x100? (multiple 0x3F800000 / 1.0f depending on the number of channels) */
 
-    /* test known versions (later versions start from 0x24 instead of 0x20) */
-    if (!(read_u32(0x38,sf) == 0x3F800000 ||
+    /* test known "version" (some float) later versions start from 0x24 instead of 0x20 */
+    if (!((read_u32(0x38,sf) == 0x3F800000 && read_u32(0x38,sf) != 0x4530F000) || /* Avengers */
```
kode54, line 52: @bnnm is this correct? You effectively test the same file offset for two different values, checking that it's equal to one and also not equal to another.

PR #728, opened 2020-10-10
wady wav

bnnm:
- Add .way bgm decoder/meta [Eien no Owari ni (PC)]
- Add .wav DSP/OPUS [Dragon Quest (Switch)]


kddlb:
@bnnm I merged anyways, since the PRs not building on CI/CD are my fault; I forgot that GHA doesn't expose secrets when building PRs.

PR #733, opened 2020-10-17
Fix build with MinGW-w64

CyberBotX:
Only works with 32-bit, however

(Sorry about the whitespace changes, my text editor removes trailing whitespace...)


kode54:
32 bit is fine for now anyway, as bundled dependencies are only included for 32 bit, and all of the Windows players only support 32 bit at this point.

PR #772, opened 2020-12-02
switch nibbles on OKI4s mono, add to TXTH

ahigerd:
It doesn't seem that OKI4s was tested on any mono samples because the games we'd looked at in the past didn't use them. Gitadora apparently does, and it turns out that the nibbles are swapped.

Gitadora also uses headerless samples for keysounds, so this also adds support for it to TXTH.

(I'm CodaHighland on Discord.)

### DIFF
--- a/src/meta/txth.c
+++ b/src/meta/txth.c
@@ -39,6 +39,7 @@ typedef enum {
     TGC = 29,           /* Tiger Game.com 4-bit ADPCM */
     ASF = 30,           /* Argonaut ASF 4-bit ADPCM */
     EAXA = 31,          /* Electronic Arts EA-XA 4-bit ADPCM v1 */
+    OKI4S = 32,         /* OKI ADPCM with 16-bit output (unlike OKI/VOX/Dialogic ADPCM's 12-bit) */
 } txth_type;
 
 typedef enum { DEFAULT, NEGATIVE, POSITIVE, INVERTED } txth_loop_t;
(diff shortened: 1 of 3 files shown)


kode54:
Nice catch.

kode54 (approved):
Looks good

src/meta/txth.c
```diff
@@ -39,6 +39,7 @@ typedef enum {
     TGC = 29,           /* Tiger Game.com 4-bit ADPCM */
     ASF = 30,           /* Argonaut ASF 4-bit ADPCM */
     EAXA = 31,          /* Electronic Arts EA-XA 4-bit ADPCM v1 */
+    OKI4S = 32,         /* OKI ADPCM with 16-bit output (unlike OKI/VOX/Dialogic ADPCM's 12-bit) */
```
NicknineTheEagle, line 42: Doc mismatch with TXTH.

kode54: Yeah, I *just* fixed that.

NicknineTheEagle:
Should haved changed nibble order for stereo, too, it's unlikely that it's different for stereo and mono.

bnnm:
I checked original DLL again and stereo does use high nibble first too (L=high), whoops (code ANDs low first but uses it later, that misguided me).

src/meta/txth.c
ahigerd: Oh, oops, thanks. I forgot to fix that comment when I added it to the doc; it was initially a quick tweak to see if it even worked.

PR #857, opened 2021-05-07
Handle The Sims 2 XA header variant, part 2

mbc07:
With previous PR I had checked only the XAs from the base game and the Expansion Packs, turns out there's another XA variant used in some of the Stuff Packs. With this PR vgmstream should **really** cover all XAs from The Sims 2, including all DLC content ever released for it. 

I also throwed the .package files that contain those XA tracks in other tools that can handle the packed format (SimPE, iLive Reader and DatGen) and found they categorize the ones with `XA\x12` magic as music, while the `XA\0\0` ones are categorized as sound. This seems to align perfectly with the regular `XAI\0` / `XAJ\0` from Sim City 3000 that vgmstream already supported, so I've updated the comments once again.


NicknineTheEagle:
Something tells me that that there's a meaning to those two bytes at 0x02. It looks like a 16-bit integer. Maybe it's flags?

mbc07:
Could be, but have no idea of what they might be flagging. AFAICT the XAs from The Sims 2 doesn't have loop points (they loop back to back), are either 22 kHz or 32 kHz and mostly mono, although there are stereo files too. Those combinations are used interchangeably with both headers, without any particular pattern (the only thing I didn't find was a stereo XA with 22 kHz sample rate using the  `XA\0\0` header). 

From the tools I mentioned, they seem to define if the file is sound or music from an additional GroupID found in each index entry of the .package file, but the actual XA, once extracted, doesn't seem to carry any particular info about that other than these two differing bytes. The same GroupID for sound is also used interchangeably between XA and SPX (a completely different format), the only reliable way of telling them apart is comparing the magic on the header...

PR #895, opened 2021-07-16
ffmpeg version control for cmake

PredatorCZ:
Added at least some minimal form of version control for ffmpeg libraries.
Problems are ABI incompatibilities for older distros:
- Ext libs versions (built in dlls for Win): libavcodec  58, libavutil 56, libswresample 3
- Xenial uses libavcodec 56, libavutil 54, libswresample 1
- Bionic uses libavcodec 57, libavutil 55, libswresample 2
- Focal uses same vesions as Ext libs
- Stretch uses same versions as in Bionic
- Buster uses same versions as in Focal

It is impossible to build in Xenial due to missing symbols.
From what I've seen in your CI, Bionic builds fine, but still can break ABI compatibility (something to keep in mind).
I used minimum versions for Bionic.
Failure to detect correct versions will result in silent feature turn off.


kode54:
Something to be aware of: VGMStream does import its FFmpeg headers from the system, except for on Windows. For ABI compatibility, versioning C preprocessor functions should be used to pick code which is compatible with older versions, if that's even possible. You or anyone else is welcome to contribute changes which make this work there, if it's possible.

PR #896, opened 2021-07-17
CMake ffmpeg project tweaks

PredatorCZ:
As mentioned in #895 by kode, project should use system headers instead of the ones in ext_includes, eliminating possible ABI incompatibilities for unix systems. One and only way that I know of is to move ffmpeg include folders into their own sub-folder (eliminating wrong include priorities).

Another addition is possibility to build with external ffmpeg SDK, this allows use of custom ffmpeg SDK versions and x64 builds for Win64 systems.


bnnm:
Comment in CMAKE.md should go in **USE_FFMPEG** I think, and cli/Makefile probably needs `-I../ext_includes/ffmpeg` at line 21

PredatorCZ:
Thanks, for Makefile I made it only for Win, since Unix should use system dir.

kode54:
Custom is really only a problem on Windows, where we use a tiny subset of supported codecs anyway, and no video support. Full FFmpeg should technically work, but would be quite large to distribute with every plugin update.

PredatorCZ:
That is true, however custom is purely optional, it's up to user(builder) whenever he wants use vgmstream prebuilt dll, or official ffmpeg one. About redistribution, it's not required to redistribute ffmpeg libs, end user can install them as codec pack (if that's a still thing). And since I haven't found your custom ffmpeg source/repo, I cannot rebuild it for x64 code.

TLDR: Custom change does not affect current state of the project, it's only optional feature.

kode54:
I use unmodified FFmpeg source code with a cross builder script. There's a text file in the ext_libs folder that describes exactly what configuration was used.

PR #917, opened 2021-08-15
Add HITMAN Trilogy file extensions for Wwise files

Notexe:
Added in the file extensions for the HITMAN Trilogy's Wwise files. They are just normal .wem vorbis encoded files but they have custom extensions.

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -580,6 +580,8 @@ static const char* extension_list[] = {
     "wbd",
     "wd",
     "wem",
+    "wwem", //fake extension for Glacier Engine 2 WEM [HITMAN 1, 2 and 3]
+    "wwes", //fake extension for Glacier Engine 2 WEM [HITMAN 3]
     "wii",
     "wip", //txth/reserved [Colin McRae DiRT (PC)]
     "wlv", //txth/reserved [ToeJam & Earl III: Mission to Earth (DC)]
--- a/src/formats.c
+++ b/src/formats.c
@@ -580,8 +580,8 @@ static const char* extension_list[] = {
     "wbd",
     "wd",
     "wem",
-    "wwem", //fake extension for Glacier Engine 2 WEM [HITMAN 1, 2 and 3]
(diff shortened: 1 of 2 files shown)


NicknineTheEagle:
This sounds like a fake extension. Wwise is pretty standardized, you sure it's not just invented by the extractor tool you're using?

NicknineTheEagle:
AFAIK, Hitman resources aren't stored as loose files, are you sure those extensions are real?

Notexe:
Hitman stores the file types in the archive's header as metadata:
WWEM:
firefox_17Plob1gTk
WWES:
HxD_uNVNNupO5B

Which is what is used when extracting files. I believe they are just stored as .wem files before they are packed into the game but when packed they get given a filetype. MEWW (WWEM) for the WWEM files and SEWW (WWES)
HxD_seu70NzZ5O

They are loose files in a way but they do have file types which are defined in the header of the .RPKG archives.

NicknineTheEagle:
These seem more like fourCCs rather than extensions. I'm guessing WWEM stands for memory samples and WWES stands for streamed samples.

Notexe:
WWEM just stands for Wwise WEM I think.
I'm not entirely sure about WWES though (It contains all of the dialog in the game, while WWEM contains music and other sounds).

bnnm:
Those must be Hitman engine's fourCC identifiers for internal use and not output extensions.

For other formats we allow fourCC as extensions when it isn't too clear, but we have extensive knowledge of how Wwise works and it *must* load `(number).wem` (very rarely `(name).wem` or `.ogg/wav/xma` in old versions), and always enforces `.wem` extensions and would ignore others (officially confirmed by their docs).

So I think the proper course here is that the Hitman extractor tool changes .wwes/wwem to .wem, since even Wwise would reject those.

Also see [this tool and docs](https://github.com/bnnm/wwiser#wwise-engine-and-wwiser-usage) about how Wwise handles .wem and other info.

Notexe:
Fair enough, you make valid points. Closing this PR.

PR #936, opened 2021-08-31
Make and upload a static build for the Linux action

LoveEevee:
Note: building now takes longer (3.5 to 7 minutes instead of 2)


bnnm:
I think @kddlb needs to change vgmstream.org to include new artifact link though.

PR #939, opened 2021-09-03
Add hca key for Sonic Colors Ultimate

bxaimc:


ActualMandM:
thanks you blind squirrel entertainment for the encryption key ❤️

PR #952, opened 2021-09-19
Cmake: Refactor dependency fetching

LoveEevee:
- Build maiatrac3plus on targets without ffmpeg

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,22 +42,11 @@ function(set_ffmpeg onoff useForce)
   set(USE_FFMPEG ${onoff} CACHE BOOL "Use FFmpeg for support of many codecs" ${USE_FORCE_})
 endfunction()
 
-function(add_library_path lib_name static lib_path)
-	if(static)
-		add_library(${lib_name} STATIC IMPORTED)
-		set_target_properties(${lib_name} PROPERTIES IMPORTED_LOCATION ${lib_path}.a)
-	else()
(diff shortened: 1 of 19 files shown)


bnnm:
Changes are good but I think I'll set MAIATRAC3 as OFF by default even without FFmpeg.

It's a horrible thing that looks legally dangerous for vgmstream to use.

If somebody wants to enable manually it's on them but we should't promote it (I guess it was ON for emscripten but same thing, no good distributing that by default).

kddlb:
> Changes are good but I think I'll set MAIATRAC3 as OFF by default even without FFmpeg.
> 
> It's a horrible thing that looks legally dangerous for vgmstream to use.
> 
> If somebody wants to enable manually it's on them but we should't promote it (I guess it was ON for emscripten but same thing, no good distributing that by default).

Licensing issues?

bnnm:
MAIATRAC3 is, AFAIK, simply a decompiled DLL from Sony's Sound Forge, pretty much as-is. [See the code](https://github.com/emulibraries/maiatrac3plus/blob/master/MaiAT3PlusDecoder/src/MaiAT3PlusCoreDecoder.cpp) and judge.

IANAL but that doesn't look something that would legally fly as reverse engineering. and I'd rather not bring the unwanted attention of mighty corps to this puny project (unlikely as it is).

Even ignoring license issues, I find that morally unsettling and not something I'd want to use as default in vgmstream, that (even doing that it does) is reasonable reverse engineering, to me at least.

PR #961, opened 2021-09-29
msf fix ffmeg only build

PredatorCZ:
A bug was introduced in 039541f88158a691f0fdc471fe6d6ecdd2ba0481
When building only with ffmeg, causes undefined reference, since mpeg library is not being linked.

Is this a left typo? Any insignts @bnnm ?


bnnm:
`mpeg_get_samples_clean` shouldn't depend on mpeg libs (not part of ifdefs). Tried disabling mpeg and compiled fine, maybe do a full rebuild first?

PredatorCZ:
It was actually fixed in a008645c0f5165b52358f63832c173e224db521f
I had slightly out of sync upstream, my bad.
Closing pr. Sorry for disruption, haven't checked for latest changes.

PR #970, opened 2021-10-04
Improve docs

KatieFrogs:
- Improved wording in readme and build docs
- Added more links in readme
- Redone some build instructions to be more understandable
- Shifted sections around to have important ones closer to top
- Split supported formats to a separate file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,22 @@
 # vgmstream
-This is vgmstream, a library for playing streamed (pre-recorded) audio from
-video games.
+This is vgmstream, a library for playing streamed (prerecorded) video game audio.
 
 Some of vgmstream's features:
-- hundreds of video game music formats and codecs, from typical game engine files to 
-  obscure single-game codecs, aiming for high accuracy and compatibility.
-- support for looped BGM, using file's internal metadata for smooth transitions,
-  with accurate sample counts
-- subsongs, playing a format's multiple internal songs separately
-- many types of companion files (data split in multiple files) and custom containers
-- encryption keys, internal stream names, and many other unusual cases found in game audio
-- TXTH function, to add external support for extra formats (including raw audio in many forms)
-- TXTP function, for real-time and per-file config (like forced looping, removing
-  channels, playing certain subsong, or fusing together multiple files as a single one)
-- simple external tagging via .m3u files
-- plugins available for various common players and O.S.
+- [Hundreds of video game music formats and codecs](doc/FORMATS.md), from typical game engine files
+  to obscure single-game codecs, aiming for high accuracy and compatibility.
+- Support for looped BGM, using file's internal metadata for smooth transitions, with accurate
+  sample counts.
+- [Subsongs](USAGE.md#subsongs), playing a format's multiple internal songs separately.
+- Many types of companion files (data split into multiple files) and custom containers.
+- Encryption keys, internal stream names, and many other unusual cases found in game audio.
+- [TXTH](doc/TXTH.md) function, to add external support for extra formats, including raw audio in
+  many forms.
+- [TXTP](doc/TXTP.md) function, for real-time and per-file config, like forced looping, removing
+  channels, playing certain subsong, or fusing multiple files into a single one.
+- Simple [external tagging](USAGE.md#tagging) via .m3u files.
+- [Plugins](#getting-vgmstream) are available for various media player software and operating systems.
 
-Latest development is here: https://github.com/vgmstream/vgmstream/
+The main development repository: https://github.com/vgmstream/vgmstream/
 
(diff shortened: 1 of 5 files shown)


KatieFrogs:
@bnnm
> README and USAGE are distributed in zips

I feel that is something that could be fixed in the zip packaging script, all it would need to do is concat the two .md files that were split. Or just include it as well.

> fragmented docs = harder to find info

I disagee. It is much easier to find what you are looking for when the file name tells exactly what it contains. As an analogy, you do not keep all the code in a single file either, that would create an unreadable mess, same rule can be applied here. It looked like a wall of boring text appended to the bottom, it was not usage instructions, but a feature showcase instead, therefore the split. And despite that, the usage doc is still currently too long to comfortably read, and I just noticed it does not even have info on the executable flags (`vgmstream-cli -h`).

> not sure about putting CMake on top of BUILD

The thing with CMake instructions being higher, I felt that Windows users would be less incentified to manually build vgmstream, because the prebuilt binaries already just work for them. For Linux, only the CLI tool is distributed, while the vgmstream123 player and the Audacity plugin that people might want are locked behind a manual build. The build doc becomes their, as one might say, "download page" for the plugin. I was hoping that an inattentive Linux user that does not like reading would scroll past the block of text with lots of `sudo apt-get install ...` near the top and quickly realize it is for them. Otherwise the first block of code that they may see would be the one with `./make-build.sh` that does not include any libraries. They would try that and get frustrated that nothing works.

> you need a compiler before building anything

The way I see it in my head is first I configure the project with build scripts (cmake/autotools/bat,ps1,sh files), which then creates another script which interfaces with the compiler (gcc/msvc/clang). This is how the commands should be run, someone unfamiliar with building might get confused why intructions go backwards, talking about compiling, then only after that start talking about configuring.

> links

I like the way wikis link everything you might want to know more about. If something is vaguely hinted at and you are confused, you can quickly get an explanation after clicking on the link. For example, it might not be immediately clear what a "Subsong" might be to an average user when it is quickly mentioned, and that you need to pass flags to take advantage of the feature. But explaining it in detail would take too much space for just a small list of main features, therefore the link. It would be easier to maintain links if each feature got its own doc, or better, used the GitHub wiki for its intended purpose, but that is out of scope for today.

bnnm:
> I disagee. It is much easier to find what you are looking for when the file name tells exactly what it contains.

I agree you don't want gigantic files, but I don't agree many small docs is automatically more helpful. From my experience, *if* people bother to read some main guide they'll scroll/CTRL+F their way and give up (or google or ask around) if target info is not found quickly enough. More files = higher the chance they give up.

> Or just include it as well

I guess for now you can leave it separate but include in the zip scripts.

> I just noticed it does not even have info on the executable flags

Used to be included, don't think it belongs. Users will try `-h` before reading, cluttered the guide, and required syncing on every CLI change. Keep in mind I make *lots* of changes, babysitting docs eats time.

> The build doc becomes their, as one might say, "download page" for the plugin.

That's an interesting view. Main problem I see is that the CMake section is Linux only, but CMake is not (in fact that's its selling point). I dislike when projects are blunty anti-Windows/pro-whatever, so when I made BUILD I wanted to show vgmstream is reasonably neutral and welcomes all devs. I guess section needs some Windows examples.

> This is how the commands should be run.

The way I wrote BUILD was, shows new concepts step by step so that (hopefully) even a neophyte could understand what/why you need something and how vgmstream is built (b/c I'm a neophyte myself and wrote that stuff as I was learning). For example, the GCC section subtly introduces Linux's *package managers*, that are probably obvious to you, but weren't to me.

Now order is changed, so users may think: "WTF is a GCC or a MSVC? Why to they want me to install a bunch of non-explained crap? Why does it fail when I run the .ps1 like it tells me first on Windows?". You now have to jump around sections to follow the doc. Plus, I assume regular Linux users are (somewhat) used to manual compilation, and there is `./make-build-cmake.sh`, so I don't think doc needs to target experienced users.

Maybe it'd be better to add a new section with "quick examples" near top (or another doc) so you can put simple Linux steps there (though that's what the `.sh` was meant for...), and leave the long guide in regular order.

> I like the way wikis link everything you might want to know more about

I see value to link about subsongs or important sections, but (for example) is linking to BUILD every time it's mentioned really needed when you have a big "More info" section + link? Even wikis try to avoid repeating links.

> used the GitHub wiki for its intended purpose

To me important docs should go in src, as src is platform-agnostic. Github wiki is cute for semi-related stuff ([see](https://github.com/bnnm/vgmstream/wiki)) but non-standard.

Also (BUILD):
> On **Windows**, 64-bit support may work but has been minimally tested

I think this is true in any OS so it didn't say *Windows* before --can't guarantee 64-bit builds don't have subtle (mainly decoding) bugs.

PR #982, opened 2021-10-16
Add encrypted .mse and .koe

grj1234:
Same encryption method as .bgm file
.mse: SE, .koe: Voice


grj1234:
Sample files are here: [mse_koe_sample.7z](https://www75.zippyshare.com/v/7SlzTSuf/file.html) (files from Oyatsu no Jikan)

PR #1066, opened 2022-01-30
ubi_bao changes

modusc896d352:
this is an attempt to add standard "atomic bao" nomenclature for "streamed BAOs". like %08x.bao/%08x.sbao instead of (example) Common_BAO_0x5NNNNNNN, French_BAO_0x5NNNNNNN or Common_BAO_0x0NNNNNNN. this also applies to "memory BAOs" as well. without vgmstream dropping support for BAO files that are named as such.

this change was necessary due to the existence of [scimitar_new.bms](https://github.com/RetingencyPlan/le_quickbms_script_compendium/blob/master/scimitar%5Bnew_version%5D/scimitar_new.bms), which handles forge files rather differently as far as sound files are concerned. more specifically, it goes deeper into the BAO files themselves as well as aiming for a complete rewrite of scimitar.bms (now scimitar_alt.bms).

samples files can be found here.
[ac1_sample_files_with_08x_bao_naming.tar](https://krakenfiles.com/view/nlVr2CACNO/file.html)
any suggestions are welcome.

### DIFF
--- a/src/meta/ubi_bao.c
+++ b/src/meta/ubi_bao.c
@@ -135,7 +135,7 @@ static int parse_bao(ubi_bao_header* bao, STREAMFILE* sf, off_t offset, int targ
 static int parse_pk(ubi_bao_header* bao, STREAMFILE* sf);
 static VGMSTREAM* init_vgmstream_ubi_bao_header(ubi_bao_header* bao, STREAMFILE* sf);
 static STREAMFILE* setup_bao_streamfile(ubi_bao_header* bao, STREAMFILE* sf);
-static STREAMFILE* open_atomic_bao(ubi_bao_file file_type, uint32_t file_id, int is_stream, STREAMFILE* sf);
+static STREAMFILE* open_atomic_bao(ubi_bao_file file_type, uint32_t file_id, int is_stream, int stream_is_sbao, STREAMFILE* sf);
 static int find_package_bao(uint32_t target_id, STREAMFILE* sf, off_t* p_offset, size_t* p_size);
 
 static int config_bao_version(ubi_bao_header* bao, STREAMFILE* sf);
@@ -169,7 +169,7 @@ fail:
     return NULL;
 }
 
-/* .BAO - single BAO files from Ubisoft's sound engine ("DARE") games in 2008+ */
+/* .BAO - single BAO files from Ubisoft's sound engine ("DARE") games in 2007+ */
 VGMSTREAM* init_vgmstream_ubi_bao_atomic(STREAMFILE* sf) {
     ubi_bao_header bao = { 0 };
     STREAMFILE* streamData = NULL;
@@ -530,7 +530,7 @@ static VGMSTREAM* init_vgmstream_ubi_bao_sequence(ubi_bao_header* bao, STREAMFIL
 
         if (bao->is_atomic) {
             /* open memory audio BAO */
-            streamChain = open_atomic_bao(bao->cfg.file_type, entry_id, 0, sf);
+            streamChain = open_atomic_bao(bao->cfg.file_type, entry_id, 0, bao->cfg.switch_to_sbao_if_bao_is_streamed, sf);
             if (!streamChain) {
                 VGM_LOG("UBI BAO: chain BAO %08x not found\n", entry_id);
                 goto fail;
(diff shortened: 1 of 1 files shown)


bnnm:
Why is this needed instead of changing the .bms to output original names?

Games that do use the .bao/.sbao normally should define another file_type scheme when added.

modusc896d352:
i managed to gather some observations from datamining Assassin's Creed. as it turns out the engine that powers the game doesn't use names for anything at all. instead, it's using IDs and pre-calculated hashes to refer to in-game assets and the like.

say if one of the files in the game contains a pre-calculated hash like 0xb88e3eec, said hash would be followed by an ID of 0x107c0023 alongside some other numbers. then you would find the file that would contain said ID, followed by dozens of info needed to load a "BAO" asset followed by the actual asset itself. if said BAO has "multiple languages" multiple BAOs would follow.

original names can be used too for users not used to the new naming scheme, as they're (generally) easy to recognize due to how descriptive they can be, and i feel a lot of compatibility with them won't be lost with this new PR.

speaking of games that aren't Assassin's Creed, Shaun White Snowboarding uses the same "name" thing as that game does since it's kinda using the same engine but that scimitar_new.bms script supports forge files from that Snowboarding game too.

as far as other Ubisoft games that doesn't actually use the engine that powered two aforementioned games, well... i can't really say much about them at the moment. with that being said, i hope this comment clears things up.

bnnm:
That sounds reasonable enough, but:
1- you should add a small comment explaining (id).bao is used to mimic engine loading files by internal ID, and Common_x are the original filenames that are also supported.
2- streamed BAOs should only load .sbao, unless there is a good reason to load .bao (open_streamfile attempts add up extra time)

modusc896d352:
i'll see what i can do.

modusc896d352:
OK, i tried to add a comment explaining what this new change does alongside some new info. so that's number 1.
number 2 may take some time to pull off.

modusc896d352:
OK, this is already proving to be a disaster.
i'll be closing this pull request and fixing stuff through my own fork.

modusc896d352:
pull request re-opened with some improvements made to the code.

modusc896d352:
OK, as it turns out the recent PR commit only barely supports My Fitness Coach (Wii) BAO files. and even then, very few sounds are actually playable. such sounds are:
20000000 (atomic/n-mem/01-00000000/372e6d5b)
20000001 (atomic/n-mem/01-00000001/3c2ace73)
20000002 (atomic/n-mem/01-00000002/33a2dddf)
20000003 (atomic/n-mem/01-00000003/33e0a1cb)
20000004 (atomic/n-mem/01-00000004/3ceb1494)
20000005 (atomic/n-mem/01-00000005/39c3bba2)
20000006 (atomic/n-mem/01-00000006/31862f20)

that said, none of the aforementioned files work with the latest vgmstream release (the one with the "brwav-xa-ww" commit, version r1702-26-g915531b1) at all. i don't have much of an idea how to fix the rest of the sounds with this PR but i'll try.

modusc896d352:
i may have found out why vgmstream plays very few My Fitness Coach BAO files but can play most Assassin's Creed BAO files.

for version 0x1b0100 BAOs, vgmstream assigns the XMA1 codec to "codec map number 0". which is why i can play most sounds from the latter game without issue. however, if vgmstream tries to play any sound from the former that isn't any of the above, it gives up.

it may be related to a "field" right at the start of a "header BAO" file (version 0x1b0100) where a value can either be 0 or 16(0x10) depending on the platform. it resides at offset 0x24.
if said value is 16, you can see some sort of DSP-ish header right at the end of said header file. otherwise, you see an XMA1 header in its place instead.

i'm thinking about closing this pull request and opening up a separate issue about My Fitness Coach BAOs since i'm kinda out of options here. what do you think?

bnnm:
By point 2 above I meant, if there is no justification to add to an extra open_streamfile with .bao (as in, it was added "just in case"), if could be simply removed (with the benefit of a tiny performance gain). Since you added a comment that it's used in AC1 .exe then it's justified enough. You don't need to over-optimize those opens, just avoid adding more if there is no point.

Also, the new function flag isn't needed. POP2008/Shaun already set file_type == UBI_FORGE_b to signal that, and the new parameter makes `open_atomic_bao` harder to use/understand.

Because how complex those Ubi engines are it's best to open separate issues case per case.

modusc896d352:
ohhhhhhhhhhhhh.
right.

i made a new commit based on your feedback. i think i'm done with this now.

Issue #1069, opened 2022-02-04
[Bug] Build failed on macOS 12.2


xxyzz:
I use the commands from doc/BUILD.md:

```
$ mkdir build
$ cd build
$ cmake -DBUILD_AUDACIOUS:BOOL=OFF ..
-- The C compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found PkgConfig: /usr/local/bin/pkg-config (found version "0.29.2")
-- Checking for module 'libavcodec'
--   Found libavcodec, version 59.18.100
-- Checking for module 'libavformat'
--   Found libavformat, version 59.16.100
-- Checking for module 'libavdevice'
--   Found libavdevice, version 59.4.100
-- Checking for module 'libavutil'
--   Found libavutil, version 57.17.100
-- Checking for module 'libavfilter'
--   Found libavfilter, version 8.24.100
-- Checking for module 'libswscale'
--   Found libswscale, version 6.4.100
-- Checking for module 'libpostproc'
--   Found libpostproc, version 56.3.100
-- Checking for module 'libswresample'
--   Found libswresample, version 4.3.100
Downloading libg719_decode (git)...
Downloading LibAtrac9 (git)...
Downloading celt-0061 (file)...
Downloading celt-0110 (file)...
-- Checking for module 'jansson>=2.3'
--   Found jansson, version 2.14
--  Option Summary
-- =================
--         MPEG: ON (system)
--       Vorbis: ON (system)
--       FFmpeg: ON (system)
--   MAIATRAC3+: OFF
--      G.722.1: ON (built-in)
--        G.719: ON (download)
--       ATRAC9: ON (download)
--     FSB CELT: ON (download)
--        SPEEX: ON (system)
--        LIBAO: ON (system)
--      Jansson: ON (system)
--
--         Building
-- =========================
--              CLI: ON
--     vgmstream123: ON
[...]

PR #1071, opened 2022-02-04
Cmake: Fix Celt not downloading

KatieFrogs:
Falls back to downloading through Git if it downloads an empty file (#1069)

### DIFF
--- a/cmake/FetchDependency.cmake
+++ b/cmake/FetchDependency.cmake
@@ -31,6 +31,7 @@ function(FetchDependency name)
 	if(NOT ARGS_FETCH_PRIORITY)
 		set(ARGS_FETCH_PRIORITY git svn file)
 	endif()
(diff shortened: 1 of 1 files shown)


kode54:
You could also make the links to xiph http instead of https. Their https server is misconfigured.

KatieFrogs:
@kode54 Absolutely not, I am not putting users at risk by using insecure protocols

xxyzz:
Would you consider adding macOS build and test job to GitHub Actions?

KatieFrogs:
@xxyzz Not a bad idea, unfortunately out of the scope of this particular pull request

Side note, I did attempt to make a macOS GitHub action for vgmstream before, but could not get it to compile

bnnm:
> Absolutely not, I am not putting users at risk by using insecure protocols
Non-git downloads that aren't checked vs some kind of valid hash put users at risk anyway (https or not). Though realistically...

PR #1078, opened 2022-02-15
Add macOS build GitHub Actions

xxyzz:


KatieFrogs:
Does it build with the `-DBUILD_STATIC=ON` option in cmake, if it does, it should probably use that

xxyzz:
> Does it build with the `-DBUILD_STATIC=ON` option in cmake, if it does, it should probably use that

The build fails imminently(https://github.com/xxyzz/vgmstream/runs/5198155000?check_suite_focus=true#step:7:63)...

KatieFrogs:
I think it is because cc and ranlib were not defined correctly, try applying this patch:
[0001-Cmake-Always-define-cc-and-ranlib-for-ffmpeg.patch.txt](https://github.com/vgmstream/vgmstream/files/8073044/0001-Cmake-Always-define-cc-and-ranlib-for-ffmpeg.patch.txt)
Change the extension back to *.patch and use this command:
`git am -k path/to/file.patch`

xxyzz:
Still the same error. Here is the end of `ffbuild/config.log`:

```
test_ld cc
test_cc
BEGIN /var/folders/05/x6p94j1j60v7g4jqv7ty70vc0000gn/T//ffconf.TrZBbSrX/test.c
    1	int main(void){ return 0; }
END /var/folders/05/x6p94j1j60v7g4jqv7ty70vc0000gn/T//ffconf.TrZBbSrX/test.c
/usr/bin/cc --static -c -o /var/folders/05/x6p94j1j60v7g4jqv7ty70vc0000gn/T//ffconf.TrZBbSrX/test.o /var/folders/05/x6p94j1j60v7g4jqv7ty70vc0000gn/T//ffconf.TrZBbSrX/test.c
/usr/bin/cc -o /var/folders/05/x6p94j1j60v7g4jqv7ty70vc0000gn/T//ffconf.TrZBbSrX/test /var/folders/05/x6p94j1j60v7g4jqv7ty70vc0000gn/T//ffconf.TrZBbSrX/test.o -static
ld: library not found for -lcrt0.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
C compiler test failed.
```

kode54:
Please see the ffmpeg-* scripts here:

https://github.com/losnoco/Cog/tree/main/Scripts

adjust as necessary for the FFmpeg features which are needed, and also to build static instead of dynamic, and also to control where they are installed and later merged.

xxyzz:
mac's gcc(clang) doesn't support the `-static` flag, use this flag will get exactly the same link error in `ffbuild/config.log`.

PR #1079, opened 2022-02-15
Allow link to continue when the flag -DUSE_MPEG=OFF is used on CMake

julianxhokaxhiu:
Fix when the flag `-DUSE_MPEG=OFF` is used on CMake

When MPEG support is disabled the struct "mpeg_frame_info" becomes unavailable, producing an error when linking, as it is required but some other code which is not protected behind the #ifdef


julianxhokaxhiu:
I identified another issue, I'm going to provide another patch ASAP.

julianxhokaxhiu:
Updated the patch, now it is the correct one, and fully works as intended.

PR #1089, opened 2022-03-06
fixed system ffmpeg detection

PredatorCZ:
Fixed system ffmpeg not being detected


KatieFrogs:
Would not FindFFmpeg.cmake file have to be fixed instead
If `FFmpeg_FOUND` does not get set for you, then that file did not find all the ffmpeg components

PredatorCZ:
You're correct, as a result I've made it more robust and added missing checks.

PR #1092, opened 2022-03-21
Bump CMake minimum version to 3.11

hcs64:
[FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) is new in 3.11. If I've misunderstood then please reject this, but it seems like 3.11 is now a minimum requirement.


bnnm:
Not sure if cmake checks are global or can be done per script, but if it's the later I think FetchConcent is only used if you don't have external libs in system, so maybe it makes more sense to do it in FetchDependency.cmake

hcs64:
Abandoning in favor of making FetchContent optional, see #1094

PR #1093, opened 2022-03-21
Further fix ktac.c without VGM_USE_FFMPEG

hcs64:
This struct can't be defined without mp4_custom_t, which requires VGM_USE_FFMPEG.

It seems like 9958c00b0e8aafa369f53975d2ed91aefd1d4fb9 was meant to fix this but missed the struct. `ktac_header_t` isn't even needed anymore since it's only used for temporary values in this one function, but this was the quickest fix.

### DIFF
--- a/src/meta/ktac.c
+++ b/src/meta/ktac.c
@@ -1,6 +1,7 @@
 #include "meta.h"
 #include "../coding/coding.h"
 
+#ifdef VGM_USE_FFMPEG
 typedef struct {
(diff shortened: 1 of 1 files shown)


hcs64:
Oops, I see this was already addressed, slightly differently (removing only the member, not the whole struct), in #1091 earlier today. Take your pick.

bnnm:
Went with the first PR to follow the order, no particular preference.

PR #1094, opened 2022-03-22
Only use FetchContent on CMake >= 3.11

hcs64:
This allows building with CMake < 3.11, dependencies won't be fetched.


hcs64:
I tested this back to CMake 3.7.3. There's an error in 3.6.3 with `VERSION_GREATER_EQUAL`, which was added in 3.7, but this was already in use.

KatieFrogs:
@hcs64 
> There's an error in 3.6.3 with `VERSION_GREATER_EQUAL`, which was added in 3.7, but this was already in use.

That could be fixed by using `NOT ${CMAKE_VERSION} VERSION_LESS "3.11.0"` instead
https://cmake.org/cmake/help/latest/command/if.html#version-comparisons

PR #1106, opened 2022-04-03
another extension add

modusc896d352:
this time i'm adding two.
the macOS version of SimCity 3000 doesn't use Maxis' XA format. instead, it uses the so-called "standard" AIFC format.
in case anyone wants to test this new change, here are the files.
[simcity3000_mac_aifc.tar](https://www.sendspace.com/file/6y932b)


modusc896d352:
OK, i did my part. now here's some unsupported AIFF files.
[[simcity_3000][mac]unsupported_aiff_files.tar](https://www.sendspace.com/file/ksf25u)
i simply can't do much here. sorry.

PR #1137, opened 2022-05-09
Fix metadata reading for some .at3 [Tales of Xillia (PS3)]

uyjulian:


bnnm:
Which files need this? I tested Xillia and all original files seem to play correctly.

uyjulian:
The files play correctly, but the loop metadata is missing.

bnnm:
Original data (.nub with RIFF) does play with loop info. I think the problem is that you may be using a rip floating around that was incorrectly ripped with extra bytes (not handling the RIFF size correctly).

These kind of faulty .at3 rips are allowed without loops rather than breaking b/c there are *a lot* of old rips like those around, but I'm trying to leave RIFF validations tight to detect them (if a ripper can't even get RIFF sizes correctly chances are the rip isn't very trustable and may have other missing metadata, for example some files have 0x0c extra bytes rather than 0x08).

So the proper course of action here would be fixing the rip by removing the extra bytes.

kode54:
VGMStream has a logging message handler, but as far as I know, it's only supported in a single threaded manner, and is only implemented in debug builds? It may be useful for this debug logging to be hooked into the console of foobar2000. Pretty much every other player that implements it doesn't have a handy logging facility, except for Linux players having stdout/stderr, and Cog on macOS having debug-only and release-also logging to Console.app feed.

bnnm:
@uyjulian There is already a log about size, but (since nobody reads them I guess) changed the meta description to clue on this.

As for reading metadata even with bad size: most of the incorrect at3 around don't loop (videos and whatnot) and aren't affected, and I want people to notice and fix the few broken looped rips. Carefully managing RIFF sizes is also useful for format detection too. So I think it's no good to allow them.

Here is a .bms script that fixes RIFFs if you need it though: https://github.com/bnnm/vgm-tools/blob/master/bms/riff.bms

@kode54 There is a log handler for non-debug messages too, see this log.h/c and [this](https://github.com/vgmstream/vgmstream/blob/master/doc/USAGE.md#logged-errors-and-unplayable-supported-files).

Needs to be compiled with VGM_LOG_OUTPUT for regular logs and VGM_DEBUG_OUTPUT for debug logs, the former is enabled in most plugins (including foobar). Player needs to handle threading when outputting to their console, but vgmstream side should be thread-safe I think.

PR #1158, opened 2022-07-01
Namco GCM patch

modusc896d352:
this is my attempt to flesh out the PS2 GCM meta code.
tested the new change against the GCM files i do have out of some Namco PS2 games.


bnnm:
Also check this for some notes in general style: https://github.com/vgmstream/vgmstream/commit/887af76ef6347173700c74994de02f13c356d0be

PR #1182, opened 2022-08-04
Added PS1 VAB format

NicknineTheEagle:


modusc896d352:
yes!!!! i've been waiting for this!

src/meta/vab.c
```diff
@@ -0,0 +1,159 @@
...
+
+    add = 0;
+    if (sfine > 15) {
+        add = 1;
+        sfine -= 16;
+    }
+
+    calc = add + (note - (center - 60));//((center + 60) - note) + add;
+    pitch = _svm_ptable[16 * (calc % 12) + (int16_t)sfine];
+    type = calc / 12 - 5;
```
belm0, line 151: I think the note 60 reference (and corresponding offset following octave divide) can be removed, and you'll find that the output is the same.

```C
interval = add + (note - center);
pitch = _svm_ptable[16 * (interval % 12) + (int16_t)sfine];
type = interval / 12;
```

src/meta/vab.c
```diff
@@ -0,0 +1,159 @@
...
+
+    sfine = fine + shift;
+    if (sfine < 0) sfine += 7;
+    sfine >>= 3;
+
+    add = 0;
+    if (sfine > 15) {
+        add = 1;
+        sfine -= 16;
+    }
```
belm0, line 147: The handing of sub-note fractions here doesn't look right.

First, note that `fine` and `shift` are unsigned range 0-127 (according to psx data format).

1. the interval between them is required to calculate output frequency scaling, so the expression should be `fine - shift` (requested fraction minus VAG center fraction), not `fine + shift`.  As coded, a VAG with non-zero center note shift will move the output frequency in the wrong direction (higher instead of lower).
2. the expression `if (sfine < 0) sfine += 7` is nonsensical because `fine + shift` will always be positive, and because the applied offset of 7 (or ~5.5 cents) has no musical meaning.

I think the correct handling would be something like:
```C
fine_interval = fine - shift;  // result range [-127..127]
add = 0;  // (better name might be note_offset)
if (fine_interval < 0) {
    fine_interval += 128;
    add = 1;
}
fine_interval >>= 3;  // quantize to 4 bits for table lookup
```

NicknineTheEagle: I'm not sure about this, I'm looking at decompiled SsPitchFromNote function in Pac-Man World and it does do addition and not subtraction, then it takes the bottom 7 bits from result and adds 128 if bit 7 is set which cannot happen because it just chopped it off.

belm0: If the PSX lib does do addition of the two 7 bit numbers (I can't think of why), certainly bit 7 can be set as a result.  In any case, the vgmstream code is not matching that because it assigns the addition to a signed int, then testing if the result is negative (which it will never be).  And nor is the vgmstream code adding 128, but 7 (which again corresponds to 5.5 cents of a music note, which doesn't make sense).

psx lib used lookup tables and all this silliness to avoid using float math.  vgmstream doesn't really need to do that, since the decoder is using floats anyway.

in a pure form, calculating the rate scale to play note X from a waveform corresponding to note Y:

```
scale = 2 ^ ((X - Y) / 12)
```
i.e. double the frequency for every octave X is higher than Y (or halve for every octave lover)

the cents or fine pitch is just the fractional part of X and Y.  For psx 7-bit form:
```
scale = 2 ^ (  (    (X + (float)X_fine/128) - (Y + (float)Y_fine/128)   ) / 12 )
```
note in the above that Y_fine (reference note) is subtracted from X_fine.

NicknineTheEagle: Read my post again, it adds fine and shift and *then* takes the bottom 7 bits from the sum.

belm0: > Read my post again, it adds fine and shift and _then_ takes the bottom 7 bits from the sum.

I understood it correctly.  That is effectively a modulo 128 after the addition.

But that is not related to `if (sfine < 0) sfine += 7`, which I've pointed out is nonsensical.  (Adding uchar's into an int will not be negative, and the magnitude 7/128 cents has no meaning in musical notation-- i.e. shifting the note by a fraction 0.055.)

NicknineTheEagle: Like I said before, Sony code has the same issue, while it does 128 like you proposed, the value it checks cannot be negative at the point where the check is done so the check just always returns false, same thing as here.

belm0: your original descirption:

> then it takes the bottom 7 bits from result and adds 128 if bit 7 is set which cannot happen because it just chopped that bit off.

I explained already, but this surmises wrong:  bit 7 (i.e. most significant bit, out of 0-7) can certainly become set after adding two 7-bit numbers.

What the psx lib is doing (add two 7-bit unsigned numbers into an 8-bit unsigned result and remove the MSB, which is a modulo of 128), is fundamentally different than what vgmstream is doing (add two 7-bit numbers into a 16-bit signed result, and check the sign i.e. bit 15).

belm0: > then it takes the bottom 7 bits from result and adds 128 if bit 7 is set

Realized I may have misread this, sorry.  If you mean `b = a & 0x7F; if (b & 0x80) {...}` then that would be bizarre...

NicknineTheEagle: I apologize, I misremembered slightly. What psx lib does is `c = a + b; c = c % 128; if (c & 0x8000) {...}`. c appears to be signed 16-bit integer and since integers are auto-promoted to 32-bit when arithmetic operation is performed, the result of that first operation cannot be negative. Then it does a modulo of 128 and checks bit 15 which can never be set unless fine value was negative.

PR #1199, opened 2022-08-12
VAB: Further work on pitch calculation

NicknineTheEagle:


src/meta/vab.c
```diff
@@ -68,10 +140,20 @@ VGMSTREAM* init_vgmstream_vab(STREAMFILE* sf) {
...
+    min_note = read_u8(entry_off + 0x06, sf);
+    max_note = read_u8(entry_off + 0x07, sf);
     wave_num = read_u16le(entry_off + 0x16, sf);
 
+    if (read_vabcfg_file(sf, program_num, tone_num, &note, &fine, &uselimits)) {
+        if (uselimits)
+            note = VAB_CLAMP(note, min_note, max_note);
+    } else {
+        note = VAB_CLAMP(60, min_note, max_note);
+        fine = 0;
```
belm0, line 152: Rather than use an arbitrary note 60, playback should prefer the center note of the VAG.  This avoids pitch shift (resampling), playing the source waveform at its natural frequency.

```C
note = VAB_CLAMP(center, min_note, max_note)
fine = (note == center && note < max_note) ? shift : 0;
```

NicknineTheEagle: No, this is deliberate because this parser is mainly intended for games which use VAB for storing SFX. Such games generally play samples with note 60 (which appears to be the default note) and rely on VAB values to set the pitch.

belm0: > this parser is mainly intended for games which use VAB for storing SFX

that sounds fine

>  Such games generally play all samples with note 60 (which appears to be the default note) and rely on VAB values to set the pitch

It's not a general rule.  The game developers may have had all kinds of reason to use other center notes, and I know of specific PS1 games that did not use 60, and in fact use a different default note for every VAG.

If 60 is the default note in the games you have in mind, hopefully they've set 60 as the center note of the VAG.  Have you checked?

NicknineTheEagle: > If 60 is the default note in the games you have in mind, hopefully they've set 60 as the center note of the VAG, and so my suggestion would still output 60.

That's the thing, the game plays all samples with note 60 and uses VAB center note to control pitch the sound is supposed to play at. I.e. they would set VAB center note higher to reduce target pitch and vice versa. Sounds weird but it is what it is.

belm0: I believe it-- there's no way to have a player default that satisfies all the warped applications.  But what you describe is specifically not the intended use of VAG center note, and I wonder what percentage of games actually do this.  It could be easily resolved by setting a single value (the 60 default note) in the vag_config scheme you have.  In contrast, trying to do that for games that use a different default note (not center note) for each VAG might be a mess.  Unfortunate...

NicknineTheEagle: It's a question of what's the most intuitive path for the end user. I wasn't really targeting normal VAB banks that store sequence samples, the goal was to target games like Spider-Man which use the previously described behavior with note 60 and they work out of the box with the current parser. I guess I could change the parser considering normal VAB are far more common but it ultimately depends on the end user's needs.

belm0: I'm also speaking of VAB's which use the waveforms as SFX, not sequences.   I have no doubt that some games did this "always play note 60 and use VAG center to control sample rate", the question is what percentage of VAB's fed to vgmstream tool do this (and specifically use note 60, not some other arbitrary note).  It's just as valid for VAB's to use a scheme like "always set VAG center to 60, and use min/max (set to the same value) to control sample rate".  Both ways allow the game code to not worry what note to play each sound at, which I believe was the intention.

Perhaps it's worth a command line option to set the playback note.  (Set the default to 60 to keep the status quo, but allow -1 to mean "user VAG center note".)

NicknineTheEagle: I was thinking of adding a parameter to vab config to use center note, yeah.

NicknineTheEagle: Added this here in https://github.com/vgmstream/vgmstream/pull/1216.

PR #1232, opened 2022-09-24
FSB5 patch.

modusc896d352:
it adds two extensions from existing FSB5 meta code.
PR is currently in "draft" mode because i have yet to test it.


modusc896d352:
tested the FSB5 .wav files with test.exe and they decode fine.
but foobar2000 component of vgmstream doesn't seem to work particularly well with it, so these files need their extension renamed to lwav.

bnnm:
Please configure your IDE/git so they use \n (LF) and that doesn't change the whole fsb5.c file to \r\n (CRLF)

modusc896d352:
i'm afraid this is going to take a while.

modusc896d352:
uhhh something bad just happened, will have to close this PR for now.

PR #1234, opened 2022-09-25
fsb5 xbb+xsb

modusc896d352:
i had to use a new branch from my vgmstream repository to make this "work", so to speak.

anyhow, changelog.

- Add .xbb+.xsb format [Gladius (Xbox)].
- Added .wav/.lwav extension to FSB5 meta [The Walking Dead: The Telltale Definitive Series (Windows 7 x64 SP1)].
- Fixed some FBS5 math related to XMA2 sounds, should have no effect on playback when playing an fsb file.


modusc896d352:
anyway, despite my best efforts i could not get FBS5 with the "wav" extension play on foobar2000 (only by renaming those files into "lwav" extension did they play there) so if there's any way i can fix this without renaming extensions please let me know. thanks.

bnnm:
Sorry, but this needs work to be mergeable:

#### if you aren't used to development don't make 1 pull request with multiple changes
Just make one PR with one each big change (fsb, then xbb). Easier for me to point out issues and for you to fix them.

#### don't write useless commit messages
*"WHAT", "last fsb5 commit", "fsb5.c" "xbb+xsb v3"*? Messages are info for everybody, not just you, and should have a summary of what has changed, not where.

#### commit at better points or flatten commits (rebase/squash/force push)
Don't pullute git history with tons of tiny commits, harder to review and check. Group changes together: commit only when one feature is done and tested, and/or flatten related commits with those git commands.

#### don't make commit with multiple unrelated changes
*WHAT* somehow changes fsb5 *and* adds skx.c, that is removed later? This pollutes git history and makes your changes less trustable.

#### don't put too many comments
Comments should be info not clear at a glance, and excess makes reading code harder. In `xbb_xsb.c` there is stuff like `uint32_t riff_offset; // "RIFF" offset pointer` --comment is obvious thus useless. `uint32_t xbb_flags` isn't so clear so a comment is ok, but both comment and varname make it look like flags found in the .xbb itself. Use something concise like: `// flags for internal use`, or a better variable name.

#### make sure your code makes sense and follows common conventions
  - In fsb5.c, code also allows files starting with "FSB5". Right below you check (twice?) that file starts with "RIFF", but that couldn't be possible. As for playing .wav, USAGE.md doc explains about foobar's priority.
  -  In xbb_xsb.c, offset sizes and sizes should be written in hex (`offset + 12` > `offset + 0x0c`), and chunk ids (fourCCs) should be read big endian. The general convention is reading them as BE ("fmt "), and only LE if it's really written as LE (" tmf").
  - xbb struct also has a `is_riff`/`is_wave` flag. At a glance this suggests xbb have non-riff data, but actually this flag is only loaded and used once. Instead just use `is_id32be` to test for expected fourCCs, or local variables for short-lived checks.
  - defines should go in CAPS: `#define XBB_BLAH` 

And so on (check other meta). All that may seem tiny details but makes reading your code easier as other devs don't have to constantly readjust their expectations.

#### make sure the format parsing design is appropriate
In xbb, try to avoid reading entries that aren't the target_subsong, as it needlessly slows down parsing the file. It also makes code messy as it's not clear what parts are from the current subsong or temp variables, those huge structs could be simplified.

Also, looks like the only info in .xbb is a size and a number of entries, and then all data pasted together, and you are forced to read the whole (potentially huge) .xbb. Generally it's unusual a game audio format (that needs quick access to subsongs without having to pre-read them) was designed like that. More often there is some header file (.xsb?) that has offsets to the xbb. Are you *really* sure parsing the .xbb and not the .xss/other file is the proper way here?

The format is also seems too ambiguous, as in, other bigfile of wavs pasted together could look like that. vgmstream should read audio formats, and not bigfiles hard to classify as audio formats. Preferably post samples first to discuss and verify.

***

PRs are always welcome but you also have to make it easy for the people checking them. Also understand I'm writting all this so you may improve and future submissions are smoother.

modusc896d352:
yeah, whatever, i'm closing this

PR #1366, opened 2023-06-01
Some fixes

Hexagon123:
- Add .TRK extension [Bloody Roar: Primal Fury (GC)]
- Add .FSB.PS3 extension [Tony Hawk's Proving Ground (PS3)]
- Fix some a few DSP [Hulk (GC)]
- Fix some .STR+WAV [Burger King Big Bumpin (Xbox)]
- Add .MOV extension

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -338,6 +338,7 @@ static const char* extension_list[] = {
     "mogg",
     //"m4a", //common
     //"m4v", //common
+    //"mov", //common
     //"mp+", //common [Moonshine Runners (PC)]
     //"mp2", //common
     //"mp3", //common
@@ -559,6 +560,7 @@ static const char* extension_list[] = {
     "tk5",
     "tmx",
     "tra",
+    "trk",
     "tun",
     "txth",
     "txtp",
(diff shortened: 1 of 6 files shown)


bnnm:
- my Hulk samples have coefs at 0x14, samples of files at 0x1c?
- if Big Bumpin does need 0x9000 for multitracks, better remove comment `/* assumed for multitrack */`
- when adding common exts like .mov also include them in `common_extension_list` in formats.c
- try to avoid generic commit messages like "Update ngc_adpdtk.c"

Hexagon123:
RSD4 was 0x14, RSD3 was 0x1c

Looks like we need minor changes

[files.zip](https://github.com/vgmstream/vgmstream/files/11629533/files.zip)

bnnm:
- For .rsd, must use 0x14 or 0x1c depending on version so it doesn't break other files
- For Big Bumpin, change is fine but comment misleading
- For .mov, comment in formats.c doesn't do anything, extension must be added in the 2nd list
 
See [commit](https://github.com/vgmstream/vgmstream/commit/288d1e57d02ff5882dcbfdf3f0547a59c9ddf99f).

PR #1373, opened 2023-06-12
AWD: RenderWare Audio Wave Dictionary

EdnessP:
Not a very common format to my knowledge, mostly only ever seen used in Criterion's own games alongside with the far more common RWS (Burnout 3: Takedown, Burnout Revenge, Burnout Dominator, Black).  The only other game outside of Criterion I know of that uses this is Call of Duty: Finest Hour.

Supports format IDs 0 (PS-ADPCM), 1 (PCM S16LE), 3 (DSPADPCM), 4 (Xbox IMA ADPCM).  No idea what format ID 2 is, doesn't exist in the aforementioned games.  Seems to be Mono only as well.

Also registered H/L-prefixed extensions for other already supported formats which are used for High/Low vehicle engine sounds in the PSP and Xbox 360 Burnout games.

attels

### DIFF
--- a/doc/FORMATS.md
+++ b/doc/FORMATS.md
@@ -156,6 +156,7 @@ This list is not complete and many other files are supported.
 	- .ast (GC AFC ADPCM, 16 bit PCM)
 	- .aud (IMA ADPCM, WS DPCM)
 	- .aus (PSX ADPCM, Xbox IMA ADPCM)
+	- .awd/.hwd/.lwd (PSX ADPCM, XBOX IMA ADPCM, GC DSP ADPCM, 16 bit PCM)
 	- .brstm (GC DSP ADPCM, 8/16 bit PCM)
 	- .emff (PSX APDCM, GC DSP ADPCM)
 	- .fsb/wii (PSX ADPCM, GC DSP ADPCM, Xbox IMA ADPCM, MPEG audio, FSB Vorbis, MS XMA)
--- a/src/formats.c
+++ b/src/formats.c
@@ -89,6 +89,7 @@ static const char* extension_list[] = {
     "awa", //txth/reserved [Missing Parts Side A (PS2)]
     "awb",
     "awc",
+    "awd",
 
     "b1s",
     "baf",
@@ -230,6 +231,7 @@ static const char* extension_list[] = {
     "hxg",
     "hxx",
     "hwas",
+    "hwd",
 
     "iab",
     "iadp",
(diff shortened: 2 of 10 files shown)


bnnm:
Looks good but some comments:
- `while (next_dict_entry != linked_list_offset)` seems like would loop infinitely on bad data, may need some extra stops
- `if (header_name_offset)` can't be false (always offset + 4) so maybe you wanted to check for an empty string
- not sure if same here, but [RWS](https://github.com/vgmstream/vgmstream/blob/master/src/meta/rws.c#L61) (other RenderWare format) only uses LE tag 0x0000080d and never BE even on GC/X360, if BE 0x00000809 was added here "just in case"

EdnessP:
> * `while (next_dict_entry != linked_list_offset)` seems like would loop infinitely on bad data, may need some extra stops

I agree, gonna try to think of something.  Perhaps storing the linked list previous/next addresses and checking if they line up and point back to the previous iteration's offset.

> * `if (header_name_offset)` can't be false (always offset + 4) so maybe you wanted to check for an empty string

`header_name_offset` is the output from read_u32(offset + 4) which can return 0.  Already tested with Black where it is a null pointer and has no header string. Unless you are looking at the first commit where I did actually have that mistake, in which case yes I've already noticed it and fixed it in [254c9f6](https://github.com/vgmstream/vgmstream/pull/1373/commits/254c9f6eb1888928d5ebfeb5f57be316ad6bd45e).
attels

> * not sure if same here, but [RWS](https://github.com/vgmstream/vgmstream/blob/master/src/meta/rws.c#L61) (other RenderWare format) only uses LE tag 0x0000080d and never BE even on GC/X360, if BE 0x00000809 was added here "just in case"

Yep, it is in fact BE in the GCN port of CoD:FH.
attels

bnnm:
> I agree, gonna try to think of something. Perhaps storing the linked list previous/next addresses and checking if they line up and point back to the previous iteration's offset.

Since bad data is unlikely, something like `if (total_subsongs > (some arbitrary max)) goto fail` should be enough I think

> header_name_offset is the output from read_u32(offset + 4) which can return 0. Already tested with Black where it is a null pointer and has no header string. Unless you are looking at the first commit where I did actually have that mistake, in which case yes I've already noticed it and fixed it in [254c9f6](https://github.com/vgmstream/vgmstream/pull/1373/commits/254c9f6eb1888928d5ebfeb5f57be316ad6bd45e).

Indeed I was looking at the wrong commit, sorry

EdnessP:
Done, added a check for 1024 entries which is probably overkill, *but* you never know :P Currently the highest I know of is in Black that has 105 entries.  According to research from debug symbols in a build of Burnout Revenge, the max is supposed to be 55, but either Criterion increased it in later revisions of RenderWare, or the number is supposed to be doubled, or something else, I dunno.
Also a 2nd check if the entry pointers are within the header size range and don't go somewhere outside.

PR #1376, opened 2023-06-20
VAGWAD: Jak & Daxter series

EdnessP:
While there is support in the [pre-extracted VAGs](https://github.com/vgmstream/vgmstream/blob/master/src/meta/vag.c#L85), it having to guess the interleave size and such seems a bit messy, so this can take the original VAGWAD files as input and parse the VAGDIR.AYB companion file that determines whether it's stereo among other info.

And while I see the comment `/* Jak 3 interleave in rare files, no header */` in the plain VAG parser, I personally couldn't find any subsong that had a missing 2nd stereo header - I'm chalking that up to likely a bad file extraction, since up until [last month](https://github.com/EdnessP/scripts/blob/main/jak-daxter/jak3-Xvagwad.py) there doesn't seem to have been a proper Jak 3 and Jak X VAGWAD extractor.

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -143,6 +143,7 @@ static const char* extension_list[] = {
     "ckd",
     "cks",
     "cnk",
+    "com",
     "cpk",
     "cps",
     "csa", //txth/reserved [LEGO Racers 2 (PS2)]
@@ -180,6 +181,7 @@ static const char* extension_list[] = {
     "eas",
     "eda", //txth/reserved [Project Eden (PS2)]
     "emff", //fake extension for .mul (to be removed)
+    "eng",
     "enm",
     "eno",
     "ens",
@@ -194,6 +196,7 @@ static const char* extension_list[] = {
     "filp",
     //"flac", //common
     "flx",
+    "fre",
     "fsb",
     "fsv",
     "fwav",
@@ -206,6 +209,7 @@ static const char* extension_list[] = {
     "gcub",
     "gcw",
     "genh",
+    "ger",
     "gin",
     "gms",
     "grn",
@@ -256,12 +260,14 @@ static const char* extension_list[] = {
     "isb",
     "isd",
     "isws",
+    "ita",
     "itl",
     "ivaud",
     "ivag",
     "ivb",
     "ivs", //txth/reserved [Burnout 2 (PS2)]
 
+    "jap",
     "joe",
     "jstm",
 
@@ -275,6 +281,7 @@ static const char* extension_list[] = {
     "kno",
     "kns",
     "koe",
+    "kor",
     "kraw",
     "ktac",
     "ktsl2asbin",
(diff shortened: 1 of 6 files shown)


bnnm:
On a quick look I'm not sure if this quite fits. Some considerations to discuss first:

- vgmstream may handle obscure formats/containers, but for cases that are a bit dubious (index-like 'headers'), or not quite an "audio format", TXTH may be a better fit.
 
  Question being, is this a real "audio bank" (useful in src), or a just a barely detectable bigfile-ish index (ok using TXTH).

- Generally we should try to follow how games handle audio

  Probably game opens .AYB, parses/validates it then loads VAGWADs based on lang. For bank formats we usually parse the header then open streams (or other way around for older cases but usually go in pairs). Here we can't detect lang I guess, so opening the VAGWAD first is needed then .AYB.

  That's brittle since VAGWAD are just VAGps pasted together not really distinct from its extracted data. If vag.c added the .int extension this would break; extra checks are needed in vag.c then etc, but overall feels messy for a proper format.

- .AYB does have a stereo flag (like an "audio bank"), but also hardcoded interleave and no other audio params (like an "index")

  Code is fine, so I don't want to reject it outright, but these quasi-formats are a pain to keep around/maintain (so TXTH is recommended for them, and if enough cases are found then it's useful to add to src).

***

So basically not sure what to do, also considering your .py extractor covers and documents this.

As for vag.c, guessing interleaves is ugly but plain VAG (and DSP) is special in that devs hacked around the 'official' format often, so having heuristics to autodetect/document various cases is ok.

***

Other code comments:
- try to comment out unused vars like header_name
- new metas in meta.h and vgmstream.c preferably go at the end. 
  Those parsers would be for less used formats, so by putting them in the middle, takes longer to reach common formats (small but good form)
- that Jak 3 comment in vag.c was indeed a bad extraction

EdnessP:
Yeah, funnily enough I also had that cross my mind multiple times whether or not this would really fit here (especially when it came to writing that long list of known VAGWAD extensions) so I wouldn't stress over it if it was closed.

Yes, the game does load the VAGDIR table first and has it reside in IOP memory at all times, but I originally went the route of loading it via a VAGWAD because it didn't seem like I could select which specific VAGWAD I wanted to load if I were to just load VAGDIR.AYB.  Well, other than just loading every single one in the current folder.

The interleave is hardcoded, but differs - Jak II and Jak 3 has it set to 0x2000, and Jak X has it set to 0x1000 which seems to be differentiated by the VAGDIR.AYB version number.  I guess one of my original reasons to write it was because of that comment about a bad Jak 3 .VAG file, but since now that's confirmed to be from a bad extract, eh.

bnnm:
Decided to close this, given it's such an specific case and most people would prefer extracted files. Also worth noting Jak collection uses extracted files as well, with same names as VAGWAD extracts. Hope you don't get discouraged though.

PR #1379, opened 2023-06-25
RWS: 0x809 File ID revision 

EdnessP:
Currently seems to only be known being used in Burnout 2: Point of Impact (PS2, 2002) as its only RWS format, and also alongside the common 0x80D File ID revision in the GCN & Xbox (2003) ports.

Seems to be Mono only as well like AWD (also 0x809 ID).  Platform/codec detection could probably use some work, but PS2 seemed to always use the usual PS-ADPCM, while GCN and Xbox - seemingly always signed 16-bit PCM.

Also, while this probably doesn't really matter much, wasn't entirely sure on the naming of the function and file.  Originally it was rws_809.c / init_vgmstream_rws_809, eventually I changed it to _mono.  Not really a huge fan of either name, though.

### DIFF
--- a/src/libvgmstream.vcxproj
+++ b/src/libvgmstream.vcxproj
@@ -606,6 +606,7 @@
     <ClCompile Include="meta\rsd.c" />
     <ClCompile Include="meta\rsf.c" />
     <ClCompile Include="meta\rws.c" />
+    <ClCompile Include="meta\rws_mono.c" />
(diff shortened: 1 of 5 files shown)


bnnm:
It'd probably gone with rws_old, rws_v1 or along the lines (_mono or _809 don't quite articulate the diffs at a glance) but doesn't matter that much

PR #1391, opened 2023-07-08
Sony BNK: Guesstimating sample rate from pitch

EdnessP:
May also need a lot more testing.  I was uncertain on whether to keep the hardcoded center note/fine sample rate presets from the original implementation.

Here's some data I gathered for personal research a couple days ago of the guesstimated sample rates, comparing the expected sample rates with the current implementation and the original implementation.  And also adding 1 to the center fine, and seeing by how much it differs.

| Note/Fine | Expected Hz | New Hz | New Fine+1 Hz | Old Hz | Old Fine+1 Hz |
|-----------|-------------|--------|---------------|--------|---------------|
| DC/00     | 192000      | 192000 | 192082        | 192000 | 192084        |
| C4/00     | 48000       | 48000  | 48023         | 48000  | 48021         |
| C2/44     | 44100 (?)   | 44097  | 44109         | 44048  | 44067         |
| C2/42     | 44100       | 44050  | 44074         | 44010  | 44029         |
| B6/44     | 22050 (?)   | 22042  | 22054         | 22024  | 22033         |
| B6/42     | 22050       | 22031  | 22031         | 22005  | 22014         |
| AA/44     | 11025 (?)   | 11027  | 11027         | 11012  | 11016         |
| AA/42     | 11025       | 11015  | 11015         | 11002  | 11007         |
| A4/7C     | 8000        | 7992   | 8003          | 7973   | 7977          |

### DIFF
--- a/src/meta/bnk_sony.c
+++ b/src/meta/bnk_sony.c
@@ -4,6 +4,7 @@
 #include "../util/endianness.h"
 
 typedef enum { PSX, PCM16, ATRAC9, HEVAG } bnk_codec;
+uint16_t ps_note_to_pitch(uint16_t center_note, uint16_t center_fine, uint16_t note, int16_t fine);
 
 /* .BNK - Sony's SCREAM bank format [The Sly Collection (PS3), Puyo Puyo Tetris (PS4), NekoBuro: Cats Block (Vita)] */
 VGMSTREAM* init_vgmstream_bnk_sony(STREAMFILE* sf) {
@@ -13,7 +14,7 @@ VGMSTREAM* init_vgmstream_bnk_sony(STREAMFILE* sf) {
     uint32_t stream_size, interleave = 0;
     int channels = 0, loop_flag, sample_rate, big_endian;
     int32_t loop_start = 0, loop_end = 0;
-    uint32_t center_note, center_fine, flags;
+    uint16_t center_note, center_fine, flags, pitch;
     uint32_t atrac9_info = 0;
 
     int total_subsongs, target_subsong = sf->stream_index;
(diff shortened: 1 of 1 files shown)


bnnm:
The hardcoded values (ex. b6/42 = 22050) where gotten in reverse. That is, if you give the SCREAM tool a 22050 wav, you get a b6/42 .bnk (as as per comment, you can't give it a 22140hz wav).

I assumed lib hardcodes those base values (at least on PS3/Vita/PS4 that would use software pitch), as if a sound designer includes a 44100 voice I'd be strange that it's played at an inexact 44050, but didn't test actual cases and may be a known defect.

EdnessP:
I guess this is more of an issue that it seems to be strictly 48000-base, which can't be directly translated to a 44100-base sample rate without some minor deviation.  Though PS2's SCREAM.IRX does have an oddball check and conversion for 44100-base sounds, but that's a whole 'nother thing.  (None of the samples I looked at seemed to have it, I assumed it to be some PS1 leftover)

Using its conversion (0x1000 * 44100) / 48000 results a pitch of 0xEB3 which converted to a guesstimated sample rate results in 44097 Hz - the sample rate you'd get from center note/fine C2/44 instead of C2/42.  Same with other 48000-base sounds, e.g. 24000 = 0x800 pitch: (0x800 * 44100) / 48000 => (0x759 * 48000) / 0x1000 => 22042 Hz.  This puzzled me a bit as well, which is why I wasn't really sure about it.

bnnm:
For PS2 bnk versions, (48000 * pitch) / 4096 formula would be correct as (iirc) PS2 works with hardware pitch values and it'd be standard to set the SPU2 to some pitch, even if resulting rate is not 100% exact (other PS2 formats also use pitch values like that in header).

However for PS3/Vita/etc bnks that (I assume) implement pitch in software it feels a bit odd. Well, something to keep in mind for later I guess.

EdnessP:
I should note the 44.1-base center fine discrepancy between 42 and 44 even exists in PS2 games.  Using Bully as an example again, it has plenty of C2/42, B6/42, and AA/42 center note/fine streams.  Since at a glance it seems like all of these 44.1-base sounds should actually have the fine set to 44 instead, how likely is it that Sony's tool itself just has a minor bug that sets the center fine for these streams to 42 incorrectly?

PR #1399, opened 2023-07-26
VAG: Increase bigfile padding check

EdnessP:
Fixes a whole lot of Jak & Daxter series VAGs.  Since all the different language containers use the one and the same entry table, the amount of empty space at the end will differ per language.  Originally I thought increasing the padding in the size calculation from 32KiB to 512KiB should be enough, but there still a couple sounds here and there that go way beyond that.

Like, in the Japanese localisation of Jak 1 the farmer has **52 seconds** worth of random snoring so in all the other languages, where he only has 3-6 seconds worth of snores, there's roughly 1.3 MiB of empty space at the end...


bnnm:
The check is meant to reject bigfiles with many VAGp, so other meta/TXTH have a chance to parse it. If known max padding is ~1.5MB it's fine to (say) + 0x180000 or + 0x200000 or more. If bigger padding is found then it can be increased, etc (comment is good). Doesn't need to be super accurate, just catch typical bigfiles (that would normally have +50MB of data).

EdnessP:
Alright, 2 MiB padding it is for now, just to be safe.

PR #1406, opened 2023-08-18
Handle 32-bit integer PCM in RIFF

hcs64:
That is, S.31 fixpoint.

From masters for a Sonic Origins mod.


hcs64:
Please feel free to ignore this and #1405, they were easy to add but probably not necessary.

PR #1425, opened 2023-09-27
Add support for FDK (NX) Opus Header

AGraber:
This adds support for the opus header of the Switch opus files from Nihon Falcom's FDK engine (first seen on the Switch version of Ys X -NORDICS-).

The header still has a huge chunk of unknown information but all the info for looping and decoding is read properly, and the data itself is the same format as other Switch games, so it's just a matter of decoding the header properly.

### DIFF
--- a/doc/FORMATS.md
+++ b/doc/FORMATS.md
@@ -1124,10 +1124,6 @@ different internally (encrypted, different versions, etc) and not always can be 
   - Entergram NXA header [*NXA*]
   - *opus_nxa*: `.nxa`
   - Codecs: Opus
-- **fdk_nxopus.c**
-  - Nihon Falcom FDK Opus Header [*FDK_NXOPUS*]
-  - *opus_fdk*: `.nxopus`
-  - Codecs: Opus
 - **pc_ast.c**
   - Capcom AST (PC) header [*PC_AST*]
   - *pc_ast*: `.ast`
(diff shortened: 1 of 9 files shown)


AGraber:
Sample files:

https://drive.google.com/drive/u/2/folders/1vYyC457oZc6FselEy5KYSnmOEPlIhDji

bnnm:
Looks good, but for consistency I think it's better to call it by header ID, that is: `nxof.c`, `init_vgmstream_nxof` and so on. Not a hard rule but I feel it's the cleanest way to call and identify a format (when hex-inspecting one would try to find that id first, plus Falcom could decide to reuse this format in other engines).

(files without a clean header id, may be called by extension/engine/company/etc case by case).

Similarly in `meta.h` those grouped together mean they are part of the same `.c`, new ones typically go at end, and in `vgmstream.c` new defs go before "lower priority metas".

May be useful to include `skip = switch_opus_get_encoder_delay(start_offset, sf);` in the FFMPEG #if

AGraber:
Yeah, I was initially unsure of what to name it, thanks for the heads up. I'll fix it shortly.

Regarding the grouping in meta.h, `init_vgmstream_opus_nxa` is in a similar situation than nxof (switch opus with a completely custom header grouped with the others from opus.c), which is why I decided to group it, but if you think it should be put separate then I'll do it.

Also, thanks for the tip regarding `switch_opus_get_encoder_delay`, I overlooked it from other encoders, I'll add it.

bnnm:
Thanks, this stuff evolves over time so it's not always super consistent for older files. Not a big deal anyway but might as well do it for new files.

`opus_nxa` was originally part of `opus.c`, but moved to its own file and wasn't renamed. Should be `nxa1.c`, `init_vgmstream_nxa1`, etc, I'll do it later.

(`opus.c` metas generally should only handle formats that paste a simple header over Nintendo's standard .opus header)

AGraber:
Alright, should be good to go

PR #1462, opened 2023-12-26
RWS 0x80D: DSP-ADPCM initial sample history data

EdnessP:
The field is empty in most of the GCN games I have on hand that use RWS 0x80D.  Call of Duty: Finest Hour is one such game that has non-zero data for initial sample history.

Also added a check for AWD and RWS 0x809 if the bank name is identical to the file name, and skip it from being displayed.  Consistent behavior with RWS 0x80D.

### DIFF
--- a/src/meta/awd.c
+++ b/src/meta/awd.c
@@ -6,7 +6,7 @@
 /* AWD - Audio Wave Dictionary (RenderWare) */
 VGMSTREAM* init_vgmstream_awd(STREAMFILE* sf) {
     VGMSTREAM* vgmstream = NULL;
-    char header_name[STREAM_NAME_SIZE], stream_name[STREAM_NAME_SIZE];
+    char file_name[STREAM_NAME_SIZE], header_name[STREAM_NAME_SIZE], stream_name[STREAM_NAME_SIZE];
     int /*bit_depth = 0,*/ channels = 0, sample_rate = 0, stream_codec = -1, total_subsongs = 0, target_subsong = sf->stream_index;
     int interleave, loop_flag;
     off_t data_offset, header_name_offset, misc_data_offset, linked_list_offset, wavedict_offset;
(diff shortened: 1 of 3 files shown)


bnnm:
Most of my stereo samples look like this (at offset + 0x40): d8ffd8ff 00000000, 010D010D 00000000, 00000013 00000000, 0000006D 00000000.

Repeated hists are unlikely to happen in proper DSP I think, and only having the 2nd hist (that matters less) is curious too. There is also no "initial ps" (value at offset+0x3a), and feels strange that all channels would share same hist samples. On the other hand different files have different values that do look like a sample and not garbage so it must mean something.

Maybe it's actually laid out like: hist1-l hist1-r hist2-l hist2-r (d8ffd8ff 000000 = L=d8ff,0000 R=d8ff,0000). I don't have mono or >2ch .rws to do more tests though.

EdnessP:
Here's CoD:FH's DSP header chunk data from the two files that have non-zero data, and are stereo:

**NGC_4m1.rws** (RWS 0x80D, offset 0x11CC)
```
0000:  000FF8F4 00124116  00000000 00000000
0010:  00000000 00000000  00000000 00000000
0020:  07800000 0E60F980  0C40F920 0F40F880
0030:  0A48F960 0F88F850  0DE8F9F0 00000000
0040:  000C000B 00000000  00000000 00000000
0050:  00000000 00000000  00000000 00000000
```

**NGC_4s1.rws** (RWS 0x80D, offset 0x594)
```
0000:  00250524 002A4F04  00000000 00000000
0010:  00000000 00000000  00000000 00000000
0020:  07800000 0E60F980  0C40F920 0F40F880
0030:  0A48F960 0F88F850  0DE8F9F0 00000000
0040:  056F056E 00000000  00000000 00000000
0050:  00000000 00000000  00000000 00000000
```

The initial history data in these cases would be 0x000C/0x000B and 0x056F/0x056E respectively.  RWS 0x809 and AWD also have similar DSP headers with the same peculiarities as here - most notably the coefficients being hardcoded to always be the same, and loads of empty fields like the sample rate, and of course initial predictor/scale as you've pointed out.  However unlike RWS 0x80D, each stream has its own header...  \*shrug*

Several months ago I remember checking and confirming that the initial history field is correct for AWD at the very least.  While sure, these are lossy compressions, I could still compare the same sounds from the PS2 and Xbox versions, and see if the first several samples decoded to PCM are more or less similar on GCN after applying the initial history data.  And they were.

I can't really easily reproduce this test for RWS 0x80D though, since the files in question are encoded in slightly different sample rates for each platform.  But the presence of this same data at the same location, while rare, does lead me to believe these are used.  Could maybe try and see what Dolphin does with this data, but that's for another day.

bnnm:
I think easiest to check would be editing some bgm's hists inside the iso and see how Dolphin changes-or-not the wave.

Anyway accepting it since it's a minor thing but may be worth investigating further (though initial hist in dsp-like adpcms isn't too noticeable unlike ima).

fwiw I don't think initial hist is ever set in standard .dsp headers as generated by the SDK tools even for files that don't start in silence (but it is set in other DSP-adjacent formats I think).

PR #1509, opened 2024-04-04
Add new hca key for Puyo Puyo Puzzle Pop (iOS)

sonicdemo:
I don't know what the other number next to the other keys mean, so someone'll have to fill that up, but the main key was filled up and works with the foobar2000 component.


applecuckoo:
@sonicdemo The comment is just the key but expressed in hex instead of decimal, so in your case it would be `270F`.

PR #1517, opened 2024-04-19
RAGE AUD: Additional todo note

EdnessP:
Was meant to be a part of the previous PR, but I was too slow.

Additional TODO in the main RAGE AUD meta parser about two Xbox 360 files whose block counts mismatch the actual data.  Not quite relevant yet, since XMA blocked streams are not yet supported anyway, but useful to know in the future.
- `(MC: LA) AUDIO/X360/SFX/AMBIENCE_STREAM/AMB_QUADSHOT_MALL_ADVERT_09`
- `(GTA IV) EP1_SFX/RP03_ML`

Scanned through all the audio files on PC, PS3, and X360 of both games, these seemed to be the only two outliers.

### DIFF
--- a/src/layout/blocked_rage_aud.c
+++ b/src/layout/blocked_rage_aud.c
@@ -9,7 +9,7 @@ void block_update_rage_aud(off_t block_offset, VGMSTREAM* vgmstream) {
     off_t seek_info_offset;
     int i;
 
-    //big_endian = read_u32le(0x00, sf) == 0; /* 64-bit number */
+    //big_endian = read_u32le(block_offset, sf) == 0; /* 64-bit number */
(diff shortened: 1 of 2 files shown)


EdnessP:
In AUD's case it seems like they assumed how many blocks it should take up but miscalculated it, since the final block that would be skipped over only has a tiny amount of data left over.

Here are those two files in question, in both cases it appears the final block only has a total of 0x1000 bytes of audio data.  The seek table for both indicates the data is only meant for the 2nd to last channel (one is 3ch, the other is 4ch).  The first 0x800 of audio data appear to be duplicate from a different block, but the other 0x800 for the last channel seem to be unique.

[RAGEAUD.zip](https://github.com/vgmstream/vgmstream/files/15041194/RAGEAUD.zip)

PR #1539, opened 2024-06-03
Fix build without USE_MPEG

ahigerd:
@rek in Discord discovered that building with `-DUSE_MPEG=OFF` fails because one code path doesn't properly guard mpg123 code with `#ifdef VGM_USE_MPEG`.


rek:
Wrong @rek i suspect.

ahigerd:
@rek My bad, yes, not you... I thought I'd made sure that wouldn't send a tag but I guess I made a mistake.

rek:
all good. interesting to see other projects sometimes :)

PR #1545, opened 2024-06-13
VAG: Evolution Games variant + misc

EdnessP:
- Add Evolution Games VAG variant [Rocket Power: Beach Bandits (PS2)]
- Add .RSM extension to DSP [Bully: Scholarship Edition (Wii)]
- Hide/Show SCREAM/XACT bank names if they match the input filename (like with RWS/AWD)
- Use intended AWD codec enum names


bnnm:
For enums where order matters (map 1:1 to some read int) I recommend adding the expected value: `typedef enum { VAG = 0x00, PCM = 0x01, ... }`. This makes them less brittle if moved around by mistake, and contrasts with other metas where enums are just helpers, plus you can simplify/remove the comment too.

EdnessP:
Done. Also spotted a separate bug in XWB where the stream buffer wasn't being null terminated on all get_name fail cases, causing garbage to appear in the stream name on XWBs that just have a valid bank name.
(Some rare early XWBs also seem to hang/infinitely loop(?) even on the current public build, but I haven't looked at that yet)

Issue #1548, opened 2024-06-26
[Bug] ADM3 format not recognized


zeroKilo:
Hi,
I tried the command line tool on this file but its not recognizing it:
[temp.zip](https://github.com/user-attachments/files/15987010/temp.zip)

I looked into it with a hex editor and it starts with "ADM3" (41 44 4D 33) but the tool says it failed to open it. I used the latest release, this file is from PUBG Lite, can someone help me please?

Greetz

zeroKilo:
So,
here I wrote a little tool that exports audio as raw float 32 data by using their internal "DecodeFrame" method from their dll. Still the result sounds weird. Dunno if this helps fixing your code but for anyone else its maybe a start :)
Greetz
![image](https://github.com/vgmstream/vgmstream/assets/12677118/0c37f1d3-7106-42e5-a597-a37b08aba09a)
![image](https://github.com/vgmstream/vgmstream/assets/12677118/fc83be93-4573-4ff0-9bc7-283f27cca1bd)
[REVTest2.zip](https://github.com/user-attachments/files/16010479/REVTest2.zip)

EdnessP:
Looked into it briefly, this is where the issue lies.  It's assuming the 1st unknown SMB1 table value is a 32-bit int, but turns out it's actually a 16-bit int, as the upper 2 bytes in your sample file has some non-zero values in the table.
https://github.com/vgmstream/vgmstream/blob/3df6e4303f847696b895792fd3c0025ba47489ba/src/meta/adm.c#L140-L149

zeroKilo:
oh thank you for looking into it. I currently bypass that problem by simply scanning for SMP2 tags before the data starts. My problem is, that even with the original function used to decode the frames, the resulting audio still has problems:
![image](https://github.com/vgmstream/vgmstream/assets/12677118/cc511aeb-8b9d-47d4-93f4-618182d785de)
it looks fine from zoomed out, but if I zoom in:
![image](https://github.com/vgmstream/vgmstream/assets/12677118/e2ee28d2-ff25-40e5-89a8-b246d89b9234)
you can see alot of these jumps, which result in a high pitched noise... is there any postprocessing to do after the decoding? I havent found out from your code where you actually process the data
Greetz WV

EdnessP:
[...]

PR #1549, opened 2024-06-27
Small fixes

EdnessP:
- Add XWB+XSB pairs [SpongeBob: Lights, Camera, Pants! (Xbox)]
- Fix ADM3 smp2_unk fail case for #1548 [PUBG Lite]
    - Add ADM3 GRN1 chunk support?
    (was being ignored, I don't know if on purpose / caused other bugs, nothing was mentioned in the coment)


bnnm:
For uncommon/custom xwb+xnb pairs I think it's preferable using [TXTM](https://github.com/vgmstream/vgmstream/blob/master/doc/USAGE.md#txtm) rather than keep adding more brittle/hardcoded stuff. The list existed before TXTM was added and may be trimmed down in the future (probably).

ADM3 GRN1 I think I just forgot about it. Its SMP2 seems always unique in my samples.

PR #1557, opened 2024-07-17
Add Ongakukan ADPCM decoder

modusc896d352:
Created by Ongakukan and used starting with Train Simuator Midousuji-hen on the PS2, it became their sound codec of choice for most of their PS2 and PSP games.

For various reasons, a custom decoder had to be created. Though I consider all work on this finished, it was a PITA to debug properly on VS2022 owing to some... creative decisions I had to take while making a decoder like this, plus I just couldn't debug libvgmstream project on it as I had vgmstream_cli set as the sole startup project on the whole vgmstream_full solution.

Some memory from the decoder also couldn't be freed as exception errors were plentiful while debugging in VS2022, despite the free function being called many times over.

I'm also currently aware of failed build tests happening except for Windows (Visual Studio) build and will be fixed shortly.

All of the above issues are my sole responsibility. That said, I'm currently welcoming some feedback into this one as this is my first time doing a large-scale PR like this.

### DIFF
--- a/src/base/decode.c
+++ b/src/base/decode.c
@@ -49,6 +49,10 @@ void decode_free(VGMSTREAM* vgmstream) {
         free_imuse(vgmstream->codec_data);
     }
 
+    if (vgmstream->coding_type == coding_ONGAKUKAN_ADPCM) {
+        free_ongakukan_adp(vgmstream->codec_data);
+    }
+
     if (vgmstream->coding_type == coding_COMPRESSWAVE) {
         free_compresswave(vgmstream->codec_data);
     }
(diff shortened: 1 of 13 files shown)


modusc896d352:
other test builds should no longer error out, however build logs indicate that there were still many compiler warnings that this new decoder caused. i'll address those too.

modusc896d352:
undocumented change in the commits: **libvgmstream.vcxproj.filters** had to be updated so VS2022 could recognize that **coding\libs\** folder exists by manually adding said folder to libvgmstream project, in addition i had to add in a "Header Files" folder from that same folder from before which meant i had to manually move .h files from the project to said folder. same thing was done to "Source Files" but for .c files instead.

anyway, spam over.

modusc896d352:
i tested this new decoder with new ADP files through foobar2000 (had to install a fresh fb2k-component file to install into the player tho) and i found some severe issues that will affect everyone if it isn't fixed. can't disclose them to everyone just yet, so instead i'll be closing this PR until further notice.

PR #1558, opened 2024-07-17
ongakukan-adp-v2

modusc896d352:
Adds meta+decoder combo: with this PR, vgmstream can now support **ADP** files from the following Ongakukan games:

- Train Simulator: Midousuji-hen [PS2] (2003/10/30)
- Train Simulator +(plus) Densha de Go! Tokyo Kyuukou-hen [PS2] (2003/12/18)
- Train Simulator: Kyushu Shinkansen [PS2] (2004/05/31)
- Mobile Train Simulator +(plus) Densha de Go! Tokyo Kyuukou-hen [PSP] (2005/02/17)
- Train Simulator: Keisei, Toei Asakusa, Keikyuusen [PS2] (2005/08/25)
- Mobile Train Simulator: Keisei, Toei Asakusa, Keikyuusen [PSP] (2006/02/23)

these ADP files house an audio codec that i've never seen documented anywhere on the internet. as i figured out the details of what this codec is, it turned out to be a simple 4-bit ADPCM codec that frankly isn't much to write home about. i did end up writing a custom decoder out of it though, but if nothing else i'm actually quite proud of the end result here.

in any case, this is my second large-scale PR. my first PR just went and crashed to the deepest end of the seas below; it was a mess, to put it lightly. most of this PR's work has been carried forward from the first one.

while at it, i also re-organized the **libvgmstream.vcxproj.filters** so some files and folders can finally be recognized. i added **coding\libs** filters so they also have **Header Files** and **Source Files** respectively.
i had to move .h files from the **src\coding\libs** folder to the **coding\libs\Header Files** filter.
and moved .c files from the **src\coding\libs** folder to the **coding\libs\Source Files** filter. as well.

all in all, that'll be it from me. feedback (good or bad) is welcome.

### DIFF
--- a/src/base/decode.c
+++ b/src/base/decode.c
@@ -49,6 +49,10 @@ void decode_free(VGMSTREAM* vgmstream) {
         free_imuse(vgmstream->codec_data);
     }
 
+    if (vgmstream->coding_type == coding_ONGAKUKAN_ADPCM) {
+        free_ongakukan_adp(vgmstream->codec_data);
+    }
+
     if (vgmstream->coding_type == coding_COMPRESSWAVE) {
         free_compresswave(vgmstream->codec_data);
     }
@@ -147,6 +151,10 @@ void decode_seek(VGMSTREAM* vgmstream) {
 
     if (vgmstream->coding_type == coding_IMUSE) {
         seek_imuse(vgmstream->codec_data, vgmstream->loop_current_sample);
+    }
+
+    if (vgmstream->coding_type == coding_ONGAKUKAN_ADPCM) {
+        seek_ongakukan_adp(vgmstream->codec_data, vgmstream->loop_current_sample);
     }
 
     if (vgmstream->coding_type == coding_COMPRESSWAVE) {
@@ -252,6 +260,10 @@ void decode_reset(VGMSTREAM* vgmstream) {
 
     if (vgmstream->coding_type == coding_IMUSE) {
         reset_imuse(vgmstream->codec_data);
+    }
+
+    if (vgmstream->coding_type == coding_ONGAKUKAN_ADPCM) {
+        reset_ongakukan_adp(vgmstream->codec_data);
     }
 
     if (vgmstream->coding_type == coding_COMPRESSWAVE) {
@@ -522,6 +534,8 @@ int decode_get_samples_per_frame(VGMSTREAM* vgmstream) {
             return 0; /* varies per mode */
         case coding_IMUSE:
             return 0; /* varies per frame */
+        case coding_ONGAKUKAN_ADPCM:
+            return 0; /* actually 1. */
         case coding_COMPRESSWAVE:
             return 0; /* multiple of 2 */
         case coding_EA_MT:
@@ -1527,6 +1541,10 @@ void decode_vgmstream(VGMSTREAM* vgmstream, int samples_written, int samples_to_
 
         case coding_IMUSE:
             decode_imuse(vgmstream, buffer, samples_to_do);
+            break;
+
+        case coding_ONGAKUKAN_ADPCM:
+            decode_ongakukan_adp(vgmstream, buffer, samples_to_do);
             break;
 
         case coding_COMPRESSWAVE:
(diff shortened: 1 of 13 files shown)


bnnm:
General tips, though I guess I could massage this after merging.

- its not clear to me why this is a external lib, could be simplified in the future (external libs in vgmstream are a pain to make/use/maintain).

- as a general rule libs/files don't free/close anything they havent alloc/open'ed themselves (error prone)
In other words, `close_streamfile` should be done in `ongakukan_adp_decoder.c` (where it's opened) and not in the lib (where it's just passed around).

- mark internal functions not defined in .h static (private).  Makes code clearer and compilers can do extra optimizations.
`uint8_t read_u8_wrapper(ongakukan_adp_t* handle);` > `static uint8_t read_u8_wrapper(ongakukan_adp_t* handle);` and so on

- style is unusual, making it a bit harder to follow
Use common terms (free_all > free, boot > init, grab > get), short-ish names, bool rather than chars for flags, etc.
lib design is odd at times. Why call `init_ongakukan_adp` with `sample_has_base_setup_from_the_start=1` manually? (does setting it to 0 have some actual use?)

- the common standard is that external lib functions should start with the lib name (also helps avoiding function name clashing)
`free_all_ongakukan_adpcm` > `ongakukan_adpcm_free`, etc

- too many comments make code harder to follow.
No need to explain that `read_u8_wrapper` does read a byte and so on.

- if lib doesn't support pcm16 it should `goto fail` if adpcm flag is not set

- if hist can only be pcm16, don't use void* (that means it could be float or something). You can just use `short pcm16[2]` plus removes allocs = easier.

- the fail check in riff.c seems unnecessary and confusing. 
Seems every these .adp already fails in riff.c (bad riff size plus bad or missing "data"). Best to keep things simpler and not add anything unless you actually need it (also don't add extra variables when it can be avoided, just `goto fail` directly).

- checks in `adp_ongakukan.c` don't seem very effective
That meta should ignore any valid `.wav` renamed to `.adp`, but stuff like `(bitrate != (sample_rate * block_size)` is true for most files thus not useful to test. Seems to me that in `.adp` always has `riff_size - 0x24 == (file_size - 0x2c) * 2 * 2` (or close to that) which is a lot more restrictive check few files could pass.

- use 4 spaces instead of tabs

- remove `printf`s

modusc896d352:
1. ehh, i gave this point some thought to answer to. i wrote it mostly to make my life easier, never really gave too much thought into actually integrating it to internal vgmstream code in a way that wouldn't be outright awkward. resulting code isn't too bothered though.
2. point taken, will do.
3. point taken, will do.
4. ok... i don't really have time to explain this so that'll have to be reserved to a separate comment after this one.
5. point taken, will do.
6. point taken, will do.
7. point taken, will do.
8. point taken, will do.
9. point taken, will do.
10. lol, point taken, will do.
11. point taken, will do. not sure how to convert tabs to 4 spaces in a file without manual work tho.
12. oh no, that was literally just an oversight! removing right now.

modusc896d352:
new commit arrived that should hopefully address all points raised by @bnnm.

function names named to something more "standard", char flags to bools, useless vars removed, adp meta checks relaxed, comments pared back as well... overall a lot of code has been refined

as for point 4, i had decoder lib designed the way it was for various reasons:

- trying stuff out (experimentation); i needed to see whichever vars or pieces of code i can get away with.
- - with this in mind, i can have vars mean something useful, be it by gating code to a certain var or just have it serve as offset to some file data. really, anything goes so long as the vars mean something. if var turns out to be not that useful and mostly gets in the way of working code, i gut it and write better code instead.
- - speaking of code, i may want to write something more complex, find out what needs improvement, what needs changing, what needs simplifying, and maybe refine the code to something i'm personally OK with.
- plus, i want more control in how i decode data. at present, i have some trust issues in how vgmstream does decoding internally.
- - say i want to write an ADPCM decoder for it but don't want to "follow the standard", or that i find out that internals are so tight in how samples are decoded i'm usually better off writing a decoder lib from scratch. luckily vgmstream allows custom decoders to take place but i guess this is only reserved for complex codecs.

maybe i'll make this part of internal vgmstream code in the future, but right now i'm just not sure.

bnnm:
If its a learning exercise I guess its ok. 

As you can see its very easy to get allocs wrong wasting lots of time, and the lib itself is oddly complex for such a simple codec. There are non-obvious details that vgmstream already handles that you'd potentially need to reimplement (for example if you find a stereo file). Since I want to simplify various codec related things in the future less external libs = better, so can't guarantee this won't be simplified later.

Well, as long as it works.

PR #1566, opened 2024-07-25
ongakukan-adp-v3

modusc896d352:
fixes at least one .adp file from Midousuji-hen.

when reading "fact" chunk (for checking), chunk ID is now positioned and read after fmt_size; fact_offset was introduced for this purpose. meaning, vgmstream can now recognize "fact" chunk from being positioned at either 0x24 or 0x26 from ongakukan adp meta.

"diff" var is introduced to subtract the difference between expected_size and pcm_size. if "diff" is 0, meta can go through without issue.


modusc896d352:
closing this now, i found further issues with resulting code and i'll be solving them immediately.

PR #1573, opened 2024-08-12
treyarch-ps2-snd

modusc896d352:
sorry for the mess.

what this PR does is it adds support for Treyarch's SND format, used in the following three games:

- **[Spider-Man (2002) (PS2)]**
- **[Kelly Slater's Pro Surfer (PS2)]**
- **[NHL 2K3 (PS2)]**

SND file format is a text file consisting of various lines that give out some basic information of a sound.

here is a simplied representation of such a line:
`(stream-name) (comm1) (comm-number) (comm2) (comm-number) (comm3) (comm4)`
this is just an example, actuat line format is much more complex than that.

SND relies on VBC files, how exactly are they used depends on whether or not a sound is played either on the "spu" (uploaded to memory) or on "cd" (streamed to PS2 disc):
- spu vbc, residing in "spu" folder alongside various snd files. they were reserved for much smaller sounds, could be some foley sound or whatever. "spu" denotes that this is the cas.
- STREAM.VBC, a bigfile with raw data that must be approached from somewhere. reserved from sounds that could go for over 7 minutes apiece; denoted by "cd".

code now supports both types of files. it also does not leave out the possibility of a txtm happening somewhere, but such support is untested at present.

generally speaking, i invested a lot of time in ensuring that vgmstream handles this SND format without issue. this included using existing text helpers into the meta, which meant actual line reading for basically anything the meta required. subsong support also had to stick the landing, so now you'll see proper subsong numbers (total number of subsongs and current subsong) being reported instead of stopping after the first actual line (basically how implementing said format went before i implemented subsong support to it for real).

there is also STREAM.SND file, it's unused and currently unsupported here but it's present in all three games listed above.

i also left out two stubbed metas:

- one for Treyarch's SDX format, used in exactly one game from them:
- - **[Minority Report: Everybody Runs (PS2)]**
- - - no work was done on supporting that format.
- second stub is the **Juiced** DAT format, present in **Juiced** series of games. said stub does the bare minimum needed to load the "correct" DAT format but virtually nothing was done with it since. in its place however are various comments that i had to write that should hopefully give anyone a hint on what exactly needs to be done to flesh that out into actual vgmstream meta.

good or bad, i'm currently open for feedback.


modusc896d352:
so, i took myself some free time to actually code support for Treyarch SDX meta.

bnnm:
Since it is very time-consuming to review a +800 LOC behemoth, I'll point out issues and recheck after it evolves progressively.

What I mainly need to check is if memory/streamfiles/string handling is correct but it's hard to follow so it needs some cleanup first.


**Do not upload code that does nothing.**
If `init_vgmstream_juiced_dat` isn't ready, it shouldn't be part of this PR. It's passable if part of some related file inside `#if 0 ... #endif`, but not in a dummy `.c`, there are more appropriate ways to document formats not implemented.

**Label commits properly**
Commits descriptions need to tell what they are about. "a" and "b" are not acceptable.

**Avoid including unrelated changes**
awc.c/vgmstream.c/etc only changes whitespaces yet are part of other commits. They may look like an innocent change but I still need to stop and double check you didn't sneak a bug by mistake. Makes code review even more time-consuming.

**It's not clear at a glance or from comments if this is an actual format**
It's odd that a game would parse a text file (slow) plus the first line of STREAM.SND says `; This file is just for human reference.`. If .SND is indeed used/parsed by game code (some IRX or such) make it clearer in some comment.

Even so, vgmstream doesn't really need to support every barely usable "format" devs fart around; txth + txtp exists for a reason.

Given how odd this thing is, make clearer why this format is useful in the actual code (what's this about? is the text file actually parsed? it's just for 1 giant bigfile? multiple bank-like files?). Also, we shouldn't need to find the original PR to understand what it's all about and some example games that use it, summarize that in the code itself.


**Add more validations before starting to parse text.**
Related to the above point. Many binary files are named .snd. They may reach this odd text-only thing, which will setup tons of variables, try to read as text, then fail at some point (maybe? can't quite tell in that wall of code).

Include extra validations so this format rejects non Treyarch .snd faster. Such as ignoring .snd too big, .snd that don't start with a ";" (since it looks like all do that), and so on.


**Code stuff:**
- `snprintf(vgmstream->stream_name, 256, "%08x",ps2snd.id_);` use `STREAM_NAME_SIZE`

- "comms" is typically the short form of "communications". If it's supposed to be `commands`, either write the full word or `cmds` or something else.

- `int32_t flags`: bitflags should be stored as unsigned

- `ps2snd->snd_start_offset_set` is logically only used once so this can be removed. Don't `read_bom` either unless .snd do have BOM (add a comment if they do). Things should be done for actual reasons (BOM is mostly a Windows thing).

- `text_reader_t tr`: only used in `parse_treyarch_ps2_snd`, put it there instead of passing it around in `treyarch_ps2snd` (less places something can be used = easier to understand). There are a bunch of related vars like `ps2snd->line_len` that are only needed in `parse_treyarch_ps2_snd` too.

- `for (i = 0; i < 2; i++) ... switch (i)...`: avoid code that behaves differently based on a loop index, it's hard to follow and harder still to modify. Maybe just call some function with a flag, or just call parse_commns once where it read lines and increase `total_subsongs` while skipping subsong info that aren't from `target_subsong`.

- there is a giant code block in `if (!is_comment) { .... } ` but there is no `else` nor anything meaninful after the `if`. Just return early `if (line_is_comment(ps2snd->line)) return true;` to simplify logic/indentation and remove `bool is_comment`. Less variables = less things to consider inside giant functions.

- don't write `if (ps2snd->calc_total_subsongs_first == false)`, it's a beginner's crutch and makes code harder to read for everybody else. Just get used to `if (!ps2snd->calc_total_subsongs_first)` or `if (ps2snd->calc_total_subsongs_first)`. Rename to `if (ps2snd->is_calc_total_subsongs_first)` if you still need clarity.

- avoid tons of nested ifs like `if ((sdx_test1 & 0x80808080) == 0) if(...) (if ...)`. Just fail if value isn't what you expect.

- tokens and some variables are suffixed by random `_` for no clear reason.

- don't overtest stuff like `if (ps2snd->sf_vbc) close_streamfile(ps2snd->sf_vbc);`, close won't fail even with a NULL pointer.

- there are a bunch of var names that should be named more appropriately. ex.`temp_val5` is just used to open .vbc, so `vbc_name` would be more sensical. You can reuse that for `temp_val6`, too.

- treyarch_ps2snd has both `sf_snd` and `sf_sdx`. That makes it seem like both could be set at once, which seems not your intention. Unify them in a `sf_hdr` or such so the intent is clearer. There is a bunch of dual `snd_x`/`sdx_x` vars that are equally confusing.

- `get_string_comm_key` does a lot of manual tokenization, something that has a very high chance to be done incorrectly in C and read past strings = security issues. Probably can be vastly simplified with `sscanf`. This will need to be *very* carefully checked = even longer to review/accept. Simpler algorithms = easier for everybody.

modusc896d352:
will address those later

modusc896d352:
i don't know how to address point 2 without closing PR and re-doing all that work, and that's not even mentioning getting to point 1.

bnnm:
You could just skip point 2 and explain why you can't follow it. It's meant to help you in collaborating better with others as a general tip. If you can't make a new PR copying the current changes, I can merge-squash the commits so those messages won't carry over.

modusc896d352:
well i was getting my way around git using gitkraken.

luckily my PR code wasn't all that damaging so best case i can just clone main vgmstream repo to a new folder and bring over all treyarch ps2 code to there. or could. cloning the main repo wasn't exactly successful to me so far. "wrong index size" and all that.

modusc896d352:
OK, just today i finally brought over all treyarch ps2 sound code to a new vgmstream repo (cloned into a different folder name even).

said code from said new repo compiled without problems with latest VS2022 version (17.11.2).

still, tons of work ahead of me.

modusc896d352:
PR closed, now superseded by #1590 .

PR #1574, opened 2024-08-16
Fix processing of unmodified 3DS BCSTM streams from Camelot

gheskett:
This PR properly addresses the audio corruption issues that take place in Camelot's BCSTM format (notably with Mario Golf: World Tour). Previously, the code for this hacked around the original loop points to not-so-successfully dodge misaligned interleave issues in Mario Golf, additionally affecting other less problematic games like Mario Tennis Open as a result. I've tested this PR against every single stream from both games and have confirmed that every single one loops correctly, and while using their original loop points.

This PR contains a couple of drawbacks that may require additional consideration:
- This adds two new fields to the `VGMSTREAM` struct, as I personally didn't have any good ideas on how to implement a fix without making changes to it. Any better suggestions are welcome!
- Mario Golf: World Tour's BCSTM rip on the joshw frontend has been modified some time ago with streams that remove the problematic data (which I admittedly didn't know about before I investigated this). When/if this gets merged, that rip will need to be re-updated to use the unmodified streams.

### DIFF
--- a/src/layout/interleave.c
+++ b/src/layout/interleave.c
@@ -147,6 +147,13 @@ void render_vgmstream_interleave(sample_t * buffer, int32_t sample_count, VGMSTR
                 }
             }
 
+            if (vgmstream->broken_interleave_sample_count != 0 && vgmstream->current_sample == vgmstream->broken_interleave_sample_pos) {
+                samples_per_frame = samples_per_frame_d;
+                for (ch = 0; ch < vgmstream->channels; ch++) {
+                    vgmstream->ch[ch].offset += vgmstream->broken_interleave_sample_count / samples_per_frame;
+                }
+            }
+
             vgmstream->samples_into_block = 0;
         }
     }
--- a/src/meta/bcstm.c
+++ b/src/meta/bcstm.c
@@ -79,28 +79,30 @@ VGMSTREAM * init_vgmstream_bcstm(STREAMFILE *streamFile) {
 
     /* Camelot doesn't follow header values */
     if (is_camelot_ima) {
-        size_t blocks_to_subtract;
-        size_t samples_per_byte = 2;
+        int blocks_to_subtract;
+        size_t broken_interleave_byte_offset;
+        size_t sample_count_header = read_32bitLE(info_offset + 0x2C, streamFile);
         size_t header_block_count = read_32bitLE(info_offset + 0x30, streamFile);
         size_t header_block_size = read_32bitLE(info_offset + 0x34, streamFile);
         size_t header_last_block_size = read_32bitLE(info_offset + 0x44, streamFile);
-
         size_t data_size_file = get_streamfile_size(streamFile) - start_offset;
-        size_t sample_count_header = read_32bitLE(info_offset + 0x2C, streamFile);
-        size_t badblock_sample_diff = (data_size_file * samples_per_byte / channel_count) - sample_count_header;
+        size_t badblock_sample_diff = ima_bytes_to_samples(data_size_file, channel_count) - sample_count_header;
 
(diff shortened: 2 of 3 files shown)


bnnm:
I don't think we should be handle it this way.

TLDR: first let's try to re-demux that end garbage with a `.bms`, or reimplement Camelot stuff using a blocked layout.

***

As you may have seen from recent commits I've been cleaning up vgmstream's brittle internals, b/c the thing is nigh incomprehensible and they are getting in the way of more advanced features. Adding common fields only used in 1 case/layout should be avoided as much as possible.

Some points first:
- MG:WT (AFAIK) uses a type of "blocked" format padded with garbage data at some intervals
- MG:WT doesn't even start with a valid CSTM header but also block garbage
  - even if this gets merged "*unmodified streams*" wouldn't work
- it's possible to demux those blocks and get a correct-ish `CSTM`
  - current rip was "deblocked" using a `.bms`
- but I'm not sure if header/loop offsets are meant to be used with MG:WT blocks or not
- IIRC other Camelot 3DS games don't use those blocks

With all that in mind:
- it's possible the original *fix misaligned interleave* is actually unneeded / problematic with other games and only needed in MG:WT b/c of "block" stuff
- but maybe it was only needed in MG:WT b/c of incorrectly "deblocked" data
- so the whole `broken_interleave_sample_pos` may be just an issue of the `.bms` deblocker not removing unneeded garbage at the end
- need to test if non MG:WT games do set `broken_interleave_sample_pos`/etc

So maybe we can just fix the .bms and remove the *misligned interleave loop* thing and everything may sound ok.

Another possibility: MG:WT header was actually created/programmed assuming those blocks always exist (loop offsets depend on them). We'd need to figure out what those offsets mean in MG:WT *and* other games (in other words, understand how Camelot would have programmed them). Or maybe Camelot does its own thing and we need to handle their interleave differently vs regular `CSTM` (they hardcode interleave and other stuff, after all).

Those types of odd single-game blocks can be supported in vgmstream using a "blocked" layout. The idea being, an extra function tells vgmstream how much data/samples exists in each block. See for example `nds_hwas.c`/`blocked_hwas.c`/`layout_blocked_hwas`.

Basically it's a type of specialized `interleave` layout that can do calculations like the stuff included in `interleave.c` but isolated, which signals the format behaving differently (good) rather than shoehorning in existing common layouts (bad).

gheskett:
That's a completely fair response, as this isn't exactly the most uninvasive PR. I don't have a clear picture of what exactly went through the minds of the Camelot devs with this format, but I've noticed that the Mario Golf: Would Tour streams contain an extra few bytes in the streams that aren't accounted for in the headers. I threw in a calculation (that may or may not be accurate to how they handle this) which can detect and apply these offsets in a way that works every single time. While Mario Tennis Open doesn't have the garbage problem, those streams also do not appear to have any extra data at all, which is why this calculation works both ways.

If it makes more sense to just stick with using manually deblocked streams, then this PR can be closed. With that being said, I do still think it would be best if the loop point overrides are removed, since they're also pretty much arbitrary modifications that shouldn't be necessary with properly deblocked streams.

bnnm:
I think the most likely scenario is that we need to remove those loop point overrides and tweak the MG:WT deblocker to produce a clean .bcstm, but I'll have to recheck later.

gheskett:
Ah so I took another look at the streams straight from the source; looks like those have a bunch of their own issues as well that I didn't know about going into this. I guess I wrongly assumed that the joshw rips from ~2014 were untampered with, so the best path forward here is definitely just going to be removing loop point overrides from vgmstream and ensuring the bms script is accurate. I apologize for any additional confusion I may have contributed here.

PR #1590, opened 2024-09-06
treyarch-ps2-vbc-v2

modusc896d352:
#1573 remake.

two xmplay header files ending up in this PR is accidental. i never even touched them.


modusc896d352:
ok, stuff just went wrong and i'm still new to git so closing PR now.

PR #1591, opened 2024-09-06
treyarch-ps2-vbc-v3

modusc896d352:
#1573 remake. and this time, no other files are included (unlike #1590).

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -501,6 +501,7 @@ static const char* extension_list[] = {
     "sdp", //txth/reserved [Metal Gear Arcade (AC)]
     "sdf",
     "sdt",
+    "sdx",
     "se",
     "seb",
     "sed",
@@ -1445,6 +1446,8 @@ static const meta_info meta_info_list[] = {
         {meta_EA_SBK,               "Electronic Arts SBK header"},
         {meta_DSP_ASURA,            "Rebellion DSP header"},
         {meta_ONGAKUKAN_RIFF_ADP,   "Ongakukan RIFF WAVE header"},
+        {meta_SND_VBC,              "Treyarch SND+VBC header"},
+        {meta_SDX_VBC,              "Treyarch SDX+VBC header"},
 };
 
 void get_vgmstream_coding_description(VGMSTREAM* vgmstream, char* out, size_t out_size) {
--- a/src/libvgmstream.vcxproj
+++ b/src/libvgmstream.vcxproj
@@ -684,6 +684,7 @@
     <ClCompile Include="meta\smp.c" />
     <ClCompile Include="meta\smpl.c" />
     <ClCompile Include="meta\smv.c" />
+    <ClCompile Include="meta\snd_sdx_vbc.c" />
     <ClCompile Include="meta\sndp.c" />
     <ClCompile Include="meta\snds.c" />
     <ClCompile Include="meta\sndx.c" />
--- a/src/libvgmstream.vcxproj.filters
+++ b/src/libvgmstream.vcxproj.filters
@@ -1882,6 +1882,9 @@
     <ClCompile Include="meta\smv.c">
       <Filter>meta\Source Files</Filter>
     </ClCompile>
+    <ClCompile Include="meta\snd_sdx_vbc.c">
+      <Filter>meta\Source Files</Filter>
+    </ClCompile>
     <ClCompile Include="meta\sndp.c">
       <Filter>meta\Source Files</Filter>
     </ClCompile>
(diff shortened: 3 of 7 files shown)


modusc896d352:
changes include:
- "comms" is now "cmds", "comm" is now "cmd".
- there are now a lot of comments describing how snd and sdx formats work, but the specifics of loading a vbc file (that both formats rely on) weren't fully described yet.
- bool var uses are now less prevalent than they once were.
- added tighter checks in both snd and sdx meta.
- **treyarch_ps2snd_cmds** enum is now used more widely where it makes sense (including, but not limited to, as a return value to a function).
- line reading stuff (except for the line itself) no longer resides in the **treyarch_ps2snd** struct.
- introduced **treyarch_ps2snd_entry** struct; two instances of it found in the **treyarch_ps2snd** struct:
- - one meant to represent "current sound entry as vgmstream reads it).
- - other meant to represent "selected subsong info".
- slashed temp_var5 and temp_var6, made it so "vbc name" forming happens in one var and one var only.
- tigher checking around current subsong number, now happens after sound entry is already read instead of before.

what i didn't change:
- "get_string_cmd_key" simplification has not been done yet; not exactly sure what needs to be done here, and not sure if sscanf even fits in this case.
- STREAM.SND loading; i don't intend to remove it but i'm also fairly hesitant to do work with it now.

bnnm:
Looks better now but it'll take me some time to review.

modusc896d352:
if nothing else tho i may come back and revise the code for an upcoming commit of mine, one that hopefully shouldn't make code review a miserable affair

bnnm:
Some more points:

- `line_is_comment` can be simplified to: `return val[0] == ';'`

- Not sure what `while (!ps2snd->sf_vbc/sf_stream_vbc)` is meant to do. 
Seems like it would loop infinitely if no file is found. Also seems like it opens the streamfiles on first entry (as soon as possible), rather than after finding current subsong, which would be more natural and less bug-prone.

- `if (strcmp(ps2snd.basename, "STREAM") == 0)` seems useless if you can use `if (ps2snd.selected_entry.flags & 0x10)` to check that

- It doesn't seem like separating `sf_vbc` and `sf_stream_vbc` is useful.
Maybe it's supposed to represent that the SF is an stream, but `flags` already tell you that. Having both implies both could be used at once, but only one `sf_body` is loaded, so unifying them would be simpler and remove `close_treyarch_ps2snd_streamfiles`

- `bkup_pos` is set but never used.

- I'm not sure what the name "*_go_through" is meant to represent. Current line/offset? Parser state?
Also variables typically are named using nouns and functions using verbs.

- don't use `while (true)` (line 641) when looping over fixed totals.
Unusual and one always risks of making an infinite loop if not careful. Plus add some constant so the stop value is clearer: `if (int i = 0; i < MAX_SOUND_SOURCE_TYPES; i++)` ...

- don't make blocks with misleading indentation/end like:
``` 
if (get_string_cmd_key(&lgt, ps2snd->line, &temp_val2, &temp_val3, &temp_val4) == INT) { ps2snd->entry.volume_l = ((temp_val3 * 16383) / 100) & 0xffff;
    ps2snd->entry.volume_r = ((temp_val3 * 16383) / 100) & 0xffff; }
``` 
It's easy to miss the first (`ps2snd->entry.volume_l`) part and this style leads to subtle, hard-to-find bugs

- `tmp3_is_zero` can be local, plus name isn't appropriate.
Use something like `bool is_equals = strcmp(....) == 0;`

- `temp_val2` seems like could be called `temp_cmd` and `temp_val3` just `temp_val` or `temp_num`.
Weirder still when there is no `temp_val1`.

- `goto` should be reserved for fail conditions
Meaning don't use stuff like `if (...) goto nothing_left_to_read`, you can do `if (...) { ...; return ...; }`
Plus you could take `if (sbgt->bin_pos >= ps2snd->header_size)` outside so `sdx_is_read_in_full` becomes unnecessary.

- It's not clear from the comments if the whole commands parsing is meant to mimic the IRX or just an style of parsing.
That info makes easier to review and understand design decisions. Even so, remember vgmstream doesn't really need to reimplement audio engines, just play known audio files. One could simplify the design if all `.snd` define fields in the same order (no need to tokenize).

modusc896d352:
noted

AnonBaiter:
noted

AnonRunzes:
noted

bnnm:
This doesn't seem to be worked on plus it has conflicts with current src and can't be merged, so closing it.

Feel free to reopen/readd after fixing the issues.

Issue #1378, opened 2023-06-24
[Request] TLOU BNK/STR files not playing


Xehanort88:
PC TLOU displays bnk files which can be extracted, but which don't work with vgmstream. Is this something that can be implemented or an extraction error? Please find two examples below.

[giraffe.zip](https://github.com/vgmstream/vgmstream/files/11856387/giraffe.zip)

[doors-anims-wood-aggr.zip](https://github.com/vgmstream/vgmstream/files/11856386/doors-anims-wood-aggr.zip)

Audio files are XVAG and they work out of the box (renamed).

bnnm:
This is Sony's .bnk, but a version much higher than the ones vgmstream supports. This format is very complex so I'd need many (many) files to figure out.

Xehanort88:
As in more bnk files? Because I can provide those. And the files referenced
in the str as well.

Sent from my Samsung Galaxy smartphone and typed with big fingers on a
small touchscreen. Please excuse any typographical errors.

Il Sab 24 Giu 2023, 3:02 PM bnnm ***@***.***> ha scritto:

bnnm:
Yes more bnk files like you posted.

Xehanort88:
Thank you. Please find more.

[foli-grass-tall.zip](https://github.com/vgmstream/vgmstream/files/11857619/foli-grass-tall.zip)
[vox-inf-bloater
[vox-inf-runner-female-b.zip](https://github.com/vgmstream/vgmstream/files/11857621/vox-inf-runner-female-b.zip)
-male-a.zip](https://github.com/vgmstream/vgmstream/files/11857620/vox-inf-bloater-male-a.zip)
[vox-inf-stalker-fema
[vox-slaver-male-c.zip](https://github.com/vgmstream/vgmstream/files/11857624/vox-slaver-male-c.zip)
le-c.zip](https://github.com/vgmstream/vgmstream/files/11857622/vox-inf-stalker-female-c.zip)

bnnm:
Most .bnk should work now, except those named vox-* that work differently. .str are just text info.

EdnessP:
The vox-prefixed files should now also work.

PR #1592, opened 2024-09-07
Sony BNK: 0x23 codec fixes and ZLSD support

EdnessP:
- Moved bank/stream names to be read ASAP, per todo
- Added 0x23 PCM support
- ~~Skip 0x1A/0x23 dummy entries (mostly because it made adding ZLSD much easier)~~
- Added ZLSD chunk subsong support

Most if not all the ZLSD sounds appear to only have ~1s of actual audio, even if the duration is longer.  Guess these are intended to be as some sort of preload/cache for the real external XVAG sounds that are referenced in SBlk?

As for detecting whether it's an internal or external sound, for now the most reliable way I saw was the stream header having a pattern of 2x16-bit codec and channels/2 followed by a seemingly always constant 32-bit 0x01 which goes all the way back to v12.  And the earliest file I have on hand that contains ZLSD is v15.  For external streams it'd just read that value from the stream name of the next file, or 0x00 if it's the last one.  It may not be the "correct" way to do it, but it at least works for v35 files with both SBlk+ZLSD and ZLSD-only streams.
(You're probably meant to hash the external filenames and look it up in ZLSD by that, but eh.  There are typically way more SBlk external entries than there are ZLSD entries.  See vox-joel.bnk for example - 2516 total entries, 29 SBlk streams, 313 dummy, 2174 external references, 187 ZLSD streams)

### DIFF
--- a/src/meta/bnk_sony.c
+++ b/src/meta/bnk_sony.c
@@ -47,9 +47,6 @@ typedef struct {
 
     uint32_t start_offset;
     uint32_t stream_offset;
-    uint32_t bank_name_offset;
-    uint32_t stream_name_offset;
-    uint32_t stream_name_size;
 
     uint32_t stream_size;
     uint32_t interleave;
@@ -84,15 +81,6 @@ VGMSTREAM* init_vgmstream_bnk_sony(STREAMFILE* sf) {
     vgmstream->stream_size = h.stream_size;
 
     vgmstream->meta_type = meta_BNK_SONY;
-
-    if (h.stream_name_size >= STREAM_NAME_SIZE || h.stream_name_size <= 0)
-        h.stream_name_size = STREAM_NAME_SIZE;
-
-    /* replace this with reading into the buffer ASAP when processing tables? */
-    if (h.bank_name_offset)
-        read_string(h.bank_name, h.stream_name_size, h.bank_name_offset, sf);
-    if (h.stream_name_offset)
-        read_string(h.stream_name, h.stream_name_size, h.stream_name_offset, sf);
 
     if (h.stream_name[0]) {
         get_streamfile_basename(sf, file_name, STREAM_NAME_SIZE);
@@ -611,6 +599,7 @@ static bool process_names(STREAMFILE* sf, bnk_header_t* h) {
     int table4_entry_id = -1;
     uint32_t table4_entry_idx, table4_entries_offset, table4_names_offset;
     uint32_t entry_offset, entry_count;
+    uint32_t stream_name_offset;
 
     switch (h->sblk_version) {
         case 0x03:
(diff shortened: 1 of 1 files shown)


bnnm:
Skipping stream stuff around L482 isn't very clear, some extra comments are warranted I think. Ex. "skip ZLSD/external streams" > why wouldn't we want those here, do they point to some actual file that vgmstream already supports or something? (we do play external streams in most banks since headers are in the bank).

Skipping dummies "made adding ZLSD much easier" but some reminder would be useful (detecting when target_subsong goes past total_subsongs == zlsd or some trickery?).

Generally I think it's better to make preload/dummy entries dummy (silent) since it presents the existing bank contents more naturally. For example when handling Wwise .bnk you need to externally use exact subsong numbers, so vgmstream just shows all and marks pre-load entries as "[pre]" for convenience. It could also point you to potential dummy data I guess (say devs dummy out some header but not the data).

EdnessP:
*I was originally writing a completely different comment, but all of this suddenly became irrelevant, but here it is anyway:*

On second thought, I think the "made adding ZLSD easier" note is no longer relevant, it was mostly to do with how I originally approached it by subtracting the amount of external stream references, which fell apart when I found there are also dummy external streams there.

Basically the SBlk stream entry table points to both real streams that do exist, dummy streams (pointer being just -1), and streams that only contain a name (typically ending in .xvag) and nothing else - their stream header ends there and then.  The biggest issue is probably their real path being unpredictable, as discussed on Discord some time ago for known v12 and v15 BNKs with external streams.  Sometimes they'd be in the same folder, sometimes in a subfolder, I'm not sure where they'd exist in this case either (the samples were grabbed from #1378.)

---

~~Without knowing the hash algorithm, the preload ZLSD streams will look kinda odd if their stream name is nothing but [pre]~~ lol as I was typing this I just took another random shot in the dark and guessed it being FNV-1.
So, I can probably undo the dummy/external skipping part, and just add the hash function to load a [pre] stream.  Still unsure what to do about the uncertainty of where the .xvag files would exist though.

bnnm:
Regardless of how it's handled, mainly meant some extra comments in the code to clarify "why", since the format is rather hard to keep in the head.

For example now there is a DUMMY enum/codec but not used anymore. My first instinct was "this may be a mistake" but actually was removed on purpose, but can't quite guess it. But if I see a comment nearby saying "this is done like this b/c otherwise blah blah", well, there is that.

As for how to handle streams/dummies/etc, the typical bank cases are:
- play subsongs with preload streams as-is (ie. just 1-2 second of data) and mark them as "[pre] (name)" (this allows you to know "bank's subsong is just a preload" + you can load the actual stream manually in some other folder
- play dummy entries as silence (this allows you to see the "holes")
  - older metas may skip dummies instead, but overall I feel this is messier; it's not our job to "fix" wonky banks
- play subsongs using external streams that aren't playable otherwise by loading them in code
  - try in the same folder or some typical folder, or use a `.txtm` to let user configure them
  - may just add "not found" or such to the stream name and not fail so it's easier to use if paths aren't too standard
- external streams that are actually playable may just set stream name as the path to the external file + use dummy entry, but this isn't common I think

In this case since external streams seem always playable probably not too important to actually play them and could just be handled like dummies; not sure if loop info/etc it's unique to the `.bnk` though.

EdnessP:
Thinking about some ideas and implementations, and the most ideal one I can think also feels like it'd be the slowest one.  Since SBlk and ZLSD are two entirely separate data chunks.  If I try to look up if the current SBlk stream as it's being processed has a ZLSD entry and load it ASAP, it'd throw off every subsequent target subsong index, and that'll just get super messy.  So my alternative would be to load them as-is, and when it gets to ZLSD, scan through all the SBlk stream names again looking for a hash match.  Although, now thinking about that, how would I even get the stream name without hardcoding it for a specific SBlk version, since 0x1A/0x23 doesn't have a Table4 like the older versions... Mh
I'll experiment some more tomorrow.

EdnessP:
Okay, I've reworked it a bit.  Not exactly the biggest fan of how it's set up currently though, namely how the external init_vgmstream is done.  Any other suggestions, or will this do for now?

bnnm:
I think it's fine or at least the intent is clear enough now, may merge if you don't plan to do more changes.

RIFF_ATRAC9/XVAG_ATRAC9 is mainly for doc purposes since when adding it I wasn't sure if `.bnk` would use different subtypes for subformats (ex. 0x01/03 = RIFF_ATRAC9, 0x04=RIFF_PCM, etc), could be just RIFF/XVAG I guess. I don't think `#ifdef VGM_USE_ATRAC9` is needed other than for direct `init_atrac9` access, each `init_vgmstream_xx` should `ifdef` inside as needed.

As for `[pre]`, in bnk/acb it's suffixed at the end since it may mix regular + prefetch subsongs and it looked a bit odd as a prefix, doesn't matter much I think. Not too common either.

> Also, the ATRAC9 loop info stored with SBlk streams isn't present for the external ones, if that's what you meant. It just ends after the stream name.

By that I meant, the `.bnk` could have loop or sample_rate or other info that overrides the external stream's header. Or maybe the external stream's header values have more info and override the `.bnk`'s values (I've seen both cases in other formats). In the former case it's useful to play the stream from `.bnk` (uses bnk/overriden values rather than subfile's), in the later case it's uninteresting so not important (could just use dummy/silence).

EdnessP:
Yep, fine to merge for now.

PR #1646, opened 2024-12-22
treyarch-ps2-vbc-v4

modusc896d352:
#1591 remake.

whatever issues were raised from that PR are now effectively addressed, thoough i may have left a few of them slip through the cracks. thoughts?

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -504,6 +504,7 @@ static const char* extension_list[] = {
     "sdp", //txth/reserved [Metal Gear Arcade (AC)]
     "sdf",
     "sdt",
+    "sdx",
     "se",
     "seb",
     "sed",
@@ -1448,6 +1449,8 @@ static const meta_info meta_info_list[] = {
         {meta_DSP_ASURA,            "Rebellion DSP header"},
         {meta_ONGAKUKAN_RIFF_ADP,   "Ongakukan RIFF WAVE header"},
         {meta_SDD,                  "Doki Denki DSBH header"},
+        {meta_SND_VBC,              "Treyarch SND+VBC header"},
+        {meta_SDX_VBC,              "Treyarch SDX+VBC header"},
 };
 
 void get_vgmstream_coding_description(VGMSTREAM* vgmstream, char* out, size_t out_size) {
--- a/src/libvgmstream.vcxproj
+++ b/src/libvgmstream.vcxproj
@@ -685,6 +685,7 @@
     <ClCompile Include="meta\smp.c" />
     <ClCompile Include="meta\smpl.c" />
     <ClCompile Include="meta\smv.c" />
+    <ClCompile Include="meta\snd_sdx_vbc.c" />
     <ClCompile Include="meta\sndp.c" />
     <ClCompile Include="meta\snds.c" />
     <ClCompile Include="meta\sndx.c" />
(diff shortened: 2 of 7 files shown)


bnnm:
Looks better but I'll need time before fully reviewing it.

Some quick issues:
- don't include `printf`, use `VGM_LOG()` if needed.
Make logs/comments clearer. It says it's an "Error" but continues anyway. If I see those logs and you are not around, what does it actually mean for the parser code? What should I fix?

- simplify some logical conditions
Like `if (ps2snd->sf_vbc && !vbc_load_success)` sets `vbc_load_success = true;`, then immediately will `if (vbc_load_success) break;`. Just do `if (ps2snd->sf_vbc) break;`. It looks like `vbc_load_success` matters but doesn't.

- `get_string_cmd_key` returns a `treyarch_ps2snd_cmds` but also sets `*out_val1`
Pick either way. Rename `out_val1/2/3` which look like fake/temp names to something clearer like out_cmd, out_num or such.

- [this](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L526) doesn't look right
Probably `if (ps2snd->target_subsong < 1 || ps2snd->target_subsong > ps2snd->total_subsongs || ps2snd->target_subsong > 1000) goto fail;` is what should do. Not sure why it needs a big comment below, looks like it's avoiding some common pitfalls. It makes it seem like it does more = more time to review = worse.

- Don't use [these](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L367) `,` multi-statements in `while` (or anywhere that isn't a `for` really)
Add `if (...) break;` on top for clarity, or use `do...while` instead.

- lines [here](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L326C2-L326C40) are char arrays, so compare them to chars
`if (ps2snd->line[0] == ';')`.

- [these](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L37) post-comments with an arrow are odd
Either put the comment above, or indent so it's clearer they continue from the right-side comment.

- `return` early in places like [this](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L700), [this](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L683), [this](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L860), [this](https://github.com/vgmstream/vgmstream/blob/dc1394bef36475818e69c66cb4e1e587963d5202/src/meta/snd_sdx_vbc.c#L1061), etc.
The more indentation the harder to read, the harder to read the longer this takes to understand and review.

modusc896d352:
>   Make logs/comments clearer. It says it's an "Error" but continues anyway. If I see those logs and you are not around, what does it actually mean for the parser code? What should I fix?

nothing.

i'll just close this.

AnonBaiter:
> Make logs/comments clearer. It says it's an "Error" but continues anyway. If I see those logs and you are not around, what does it actually mean for the parser code? What should I fix?

nothing.

i'll just close this.

AnonRunzes:
> Make logs/comments clearer. It says it's an "Error" but continues anyway. If I see those logs and you are not around, what does it actually mean for the parser code? What should I fix?

nothing.

i'll just close this.

PR #1677, opened 2025-02-06
Add preliminary Nintendo Alarmo SHAA/SHSA support

gizmo4487:
This PR adds support for the SHAA audio format used by Nintendo Sound Clock: Alarmo, an alarm clock released by Nintendo in October of 2024.

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -524,6 +524,8 @@ static const char* extension_list[] = {
     "sgb",
     "sgd",
     "sgt",
+    "shaa",
+    "shsa",
     "skx",
     "slb", //txth/reserved [THE Nekomura no Hitobito (PS2)]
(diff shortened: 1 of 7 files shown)


bnnm:
If you could post some samples as doesn't seem simple to obtain/test this otherwise.

gizmo4487:
> If you could post some samples as doesn't seem simple to obtain/test this otherwise.

Samples can be found here: https://mega.nz/folder/HmgAHTQC#isNyMOphcv37QmMCmAEHmw

`FirstUse_Redc_bgm_skybonus01.shsa` is the only file that loops and `TestPiano.shsa` is the only file with a PCM16LE codec. The rest of the files use NGC_DSP and do not loop.

PR #1685, opened 2025-02-22
BAF: various fixes

EdnessP:
- BAF: Add XMA2 support [PGR4 (X360), Blur (X360)]
- BAF: Add PCM v5 support [Blur 2 (X360)]
- BAF: Fix PCM BE audio [GW:RE2 (X360), PGR4 (X360)]
- BAF: Fix XMA1 v5 loops [Blur 2 (X360)]
- BAF: Fix early v4 PS-ADPCM mono streams [The Club (PS3)]
- BAF: Display bank name if different from filename
- VAG: .MSV extension [Data Design Interactive titles (PS2)]
- Sony BNK: Fail on v3 invalid name index
---
For my own sanity, can you confirm or deny whether the XMA2 WAVE chunk metadata is completely wrong and screwed up, or if I'm just not understanding something.  [Here's a sample](https://github.com/user-attachments/files/18925363/test.zip) from PGR4 + some leftover files of testing I did.
The 1st BAF commit is where I attempted to utilise the given data.  Unlike XMA1, in XMA2 there's what looks like a seek table starting at 0x58(v4)/0x5C(v5).  The last value in there seemed to always match the loop end value for sounds that loop all the way through, so I took that as the final sample count.  However the amount of (what I presumed to be) "samples" are more than double than what actually seems to be present(?) so there's a ton of silence at the end of each stream.
In the end I just resorted to doing the same thing XMA1 does, because with raw TXTH it seemed correct.

### DIFF
--- a/src/meta/baf.c
+++ b/src/meta/baf.c
@@ -128,33 +128,15 @@ VGMSTREAM* init_vgmstream_baf(STREAMFILE* sf) {
             break;
 
         case 0x08: /* XMA1 */
+        case 0x09: /* XMA2 */
             switch(version) {
-                case 0x04: /* Project Gotham Racing (X360) */
+                case 0x04: /* Project Gotham Racing 4 (X360) */
                     sample_rate     = read_u32(header_offset+0x3c, sf);
                     channel_count   = read_u32(header_offset+0x44, sf);
                     loop_flag       =  read_u8(header_offset+0x54, sf) != 0;
                     break;
 
                 case 0x05: /* James Bond 007: Blood Stone (X360) */
-                    sample_rate     = read_u32(header_offset+0x40, sf);
-                    channel_count   = read_u32(header_offset+0x48, sf);
-                    loop_flag       =  read_u8(header_offset+0x58, sf) != 0;
-                    break;
-
-                default:
-                    goto fail;
-            }
-            break;
-
-        case 0x09: /* XMA2 */
-            switch(version) {
-                case 0x04: /* Blur (X360) */
-                    sample_rate     = read_u32(header_offset+0x3c, sf);
-                    channel_count   = read_u32(header_offset+0x44, sf);
-                    loop_flag       =  read_u8(header_offset+0x54, sf) != 0;
-                    break;
-
-                case 0x05: /* Blur 2 (X360) */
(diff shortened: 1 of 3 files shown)


bnnm:
Not sure if you mean this, but the xma header looks custom-ish and seems parsed ok. 

The seek table in doesn't look like a regular seek table, which uses samples + blocks, but a bit offset table, like the loop offset (seek tables aren't set for small streams).

Block size could be 0x8000/0x10000/0x20000 but hard to guess since data fits in 1 block (the FFFFFFFF .... FFFFFFFF parts are when blocks end).

EdnessP:
If the block size were to be larger, wouldn't that also mean the stream size is a multiple of said value?  The FF padding went up in multiples of 0x800.  Some streams had a data size of, say 0xB800.
XMA2 seemed to be fairly infrequently used by them, a majority of the audio is still XMA1, but [here's a few banks](https://github.com/user-attachments/files/18928817/blocksize.zip) from Blur and PGR4 with larger XMA2 streams ranging up to 0x1F800 in size.

bnnm:
0x10000 sounds likely. However it's just info for the seek table and not really needed unless you have a standard one. It's nice to have for doc purposes tho.

The way blocks work is, XMA has VBR frames put into 0x800 packets but may be split (the beginning of a 0x800 packet may actually contain the end bits of last frame from prev packet). Being VBR you can't seek to packet 13 (13*0x800) and know which sample it is.

So XMA has a setting to make seekable blocks. If you set it to 0x10000, it won't divide last frame at the end 0x10000 and add padding instead (0xFF for XMA2 and 0x00 for XMA1). Then it'll make a seek table of samples<>blocks, where entry N would mean seeking to N * block_size to start from sample set in the entry. Without the seek table you can't know exact sample though.

I think XMA allows any block size but it'd be wasteful to set it to low (too much padding), so common values are 0x8000/0x10000/0x20000. Last block may be smaller (still padded to 0x800).

For .baf's seek table it's probably hardcoded that entry N = bit offset X = sample (512*?) * N, but who needs to seek <0x10000 streams.

EdnessP:
Gotcha.  Yeah, I scanned through all the Blur and PGR4 banks, and across 363 sounds using XMA2, only a grand total of 16 were >=0x10000 in size.

PR #1722, opened 2025-05-20
Install libvgmstream

leo60228:
libvgmstream was previously not installed by CMake. This change allows external projects to depend on libvgmstream being installed via normal library search mechanisms.


bnnm:
Seems it's missing vgmstream-config.cmake.in

There are some changes that seem to change previous behavior in unclear ways, like `EXCLUDE_FROM_ALL` that seems to be used to install manually (always installed now?)

I think not-too-recent LTS distros don't use CMake that high, 3.10 / 3.16 / 3.22 seem typical. Since compiling CLI/v123 is more common and installing libvgmstream not so, and that (I think) LTSs stick to certain CMakes requiring a bunch of hoops to change it, it seems like a kind of sudden/annoying change.

If possible it'd be better some alt way to install that works in more versions of CMake, or that 3.23 is only required when actually installing libvgmstream.

leo60228:
There were already some references to a `BUILD_SHARED_LIBS` option. I've switched to actually implementing that (and having it remain disabled-by-default). I've also switched to using a method of installing headers that should work on older CMake versions.

PR #1723, opened 2025-05-22
src/CMakeLists.txt: add missing include(GNUInstallDirs)

MaxKellermann:
This is necessary according to
 https://cmake.org/cmake/help/latest/command/install.html

Without this, "ninja install" will attempt to create "/vgmstream".


leo60228:
Huh, I could have sworn I'd checked to see if it was included already.

PR #1725, opened 2025-05-31
XWB: XSB CRC info fix

EdnessP:
- XSB: Fix CRC-16 documentation
- XWB: Fix XMA1/2 version range

XSB (and XGS as well) in revision 0x2A and newer calculate the CRC-16/IBM-SDLC data hash from 0x12 onward - documentation here (and seemingly everywhere else) implied it should still start immediately after the hash (being 0x0A in this case) which is wrong.  Took me a bit to figure out why seemingly nothing was matching up for me.


bnnm:
Microsoft's official xact2 docs use non-hex version defines so it'd be better to keep them like that I think, plus comments also weren't changed (I used hex for custom version to make them look distinct from the official versions).

bnnm:
Prey is indeed XMA1, feel free to change it (XMA1 is somewhat compatible with XMA2 so not a big deal).
Just Cause and Blue Dragon are XMA2, XMA1 has a 0x800 superframe counter so it's easy to tell apart ([>>](https://github.com/vgmstream/vgmstream/blob/58dafedc1047986621cba78fe2b6262e147f2716/src/coding/coding_utils.c#L29)).

EdnessP:
Yeah, I was aware of the nybble counter, I didn't realise it was only present in XMA1 though. Usually what stood out to me was XMA1 streams typically starting with `08000000`

PR #1736, opened 2025-07-11
srcd

Alstruit:
- Adds Capcom RE Engine srcd.


bnnm:
- should use `vgmstream_force_loop` to add loops in subfiles instead of manual `calloc` (brittle)
  You can remove `SRCD: Invalid loop points found` checks and manually adding `vgmstream->loop_flag` and related since they are done in `vgmstream_force_loop` too.
- adds "srcd" to the format list but code doesn't seem to enforce the extension, which is strange for vgmstream
- next time please avoid unrelated changes like changing those .h to CRLF (they should be LF).

PR #1737, opened 2025-07-13
Mhwk

Alstruit:
- Adds Broderbund's Mohawk engine mhk/mhwk.


bnnm:
- please add comments to identify some game(s) that use this format for easier future maintenance, and possibly post some samples too
- in `switch (format)` it defaults to PCM8 but that looks error-prone, I'd think the format only uses certain values
- for clarity you can use `if (read_u32be(0x00, sf) != 0x4D48574B) goto fail` > `if (!is_id32be(0x00, sf, "MHWK")) return NULL`, `if (chunk_id == 0x43756523)` > `if (chunk_id == get_id32be("Cue#" ))` and so on

PR #1740, opened 2025-07-14
Mhwk v2

Alstruit:
Version 2 of Broderbund's Mohawk engine mhk/mhwk.

Games include: Myst Engine, Carmen Sandiego Word Detective

Some test case [files](https://0x0.st/s/V-qwYPxmhX7qXPq1e_M8XA/8k8M.zip).

[Logs](https://0x0.st/s/H9aNYimISugq4nVPYyvPPw/8k8a.txt).

Ref: #1737.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,6 @@
 *.bat text eol=crlf
 *.vcxproj text eol=crlf
 *.vcxproj.filters text eol=crlf
+
+#Fixes CRLF/LF conflict.
+* text=auto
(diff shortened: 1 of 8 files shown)


bnnm:
Maybe there are no stereo files, but for PCM8 instead of
```
vgmstream->layout_type = layout_none;
```
probably should use:
``` 
vgmstream->layout_type = layout_interleave;
vgmstream->interleave_block_size = 0x01;
```
as otherwise both channels would sound the same.

Alstruit:
I've yet to find a stereo test file. I added in with a small comment.

PR #1752, opened 2025-08-03
MHWK: MP2 fix

Alstruit:
Add MP2 from Riven DVD.


bnnm:
It's setting `vgmstream->num_samples = 0;` but it must be filled to work, so probably this parser isn't working ok.

Best to use `if defined(VGM_USE_MPEG)` first before FFMPEG.

PR #1753, opened 2025-08-03
Add Cyberflix Dreamfactory

Alstruit:
Games include:
- Dust 1997 (3.1/95) - (Titanic: Adventure Out of Time)
- (Disney's Math/Reading Quest with Aladdin)

[Test cases](https://0x0.st/s/Mtqj4yC8hvDpTvGjmSfqSA/87Ig.zip).

[Logs](https://0x0.st/s/WbRe5jogQR2_AZpO3XZ7uw/87IY.zip).


bnnm:
- format rejection should be improved a bit before `calloc` since extensions are rather generic. For example, files seem to start with `0x00000100`, have 0 at 0x08/0x0C and "LPPALPPA" at 0x20, which would make it more robust
- `check_extensions` is case insensitive, no need to put variations
- typo in `SEGEMENT_LIMIT `
- preferably use `sf` instead of `streamFile` for consistency

Does sample_rate change per segment? Code can be left as-is for now, but I think 'segments' code could simplified by treating it as a blocked layout, if sample rate is constant (since games typically don't change rates on the fly and repeated sample_rates are just remnants).

Some segment calcs/stuff may be simplified with `allocate_segmented_vgmstream` too by separating regular and segmented cases.

Alstruit:
No I don't think the sample rate changes and I can't test all of the games to ensure that. Nonetheless, I gone ahead and implemented all of the changes.

For the header check I think LPPALPPA would be more suited.

PR #1758, opened 2025-08-09
MHWK: Fix discontinuity

Alstruit:
Fixes popping noise near the end of the file (PCM U8).

[Samples](https://0x0.st/s/w-l5LXH1t6S75DJDSAPK5Q/8Fqk.zip).

### DIFF
--- a/src/meta/mhwk.c
+++ b/src/meta/mhwk.c
@@ -59,6 +59,55 @@ VGMSTREAM* init_vgmstream_mhwk(STREAMFILE* sf) {
     /* The actual audio data starts after the header's 20 bytes not 28 bytes */
     start_offset = current_offset + 0x14;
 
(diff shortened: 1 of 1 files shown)


bnnm:
This looks rather odd to me, games don't handle silence like this (unlikely to waste of CPU time like that, specially for old games).

More likely, the game sets a max size or samples based on some header's value, or maybe it just plays those popping noises as-is. We have re-checked a bunch of cases where audio 'pops' in vgmstream that actually pops in-game too.

Since vgmstream should play streams as-is, and there are other ways to post-process files (like .txtp) I suggest recheking what the engine/format actually does first.

There are rare cases of games skipping padding (ex. PSX-ADPCM due to how hardware works), but I don't think it's the case here.

Alstruit:
I see. I checked and yes the engine play the sound as is. Thanks for the notice.

PR #1773, opened 2025-08-30
vgmstream_cli: Fix argument parsing

Alstruit:
Dash is interpreted as a filename instead of a argument.

Use optind instead of manually handling the arguments.


bnnm:
The current behavior is meant to allow flags go before or after filenames.

These commits seem to remove that, and just stop when a non-flag is found (AKA getopt default behavior, which is what was used [before](https://github.com/vgmstream/vgmstream/commit/5355db2a798f41241dfcd32498323160ec58135c) and I find fairly annoying).

Current code seemed correct for BSD and POSIX getopt so maybe make clearer when the issue happens so it can be improved while keeping it flexible.

EDIT: should be fixed now I think while keeping the intended behavior.

PR #1812, opened 2025-12-05
Add support for '.afs' container.

Clownacy:
Tested with the 'BGM_US.AFS' and 'BGM_JP.AFS' files from Sonic Gems Collection (PS2).

The code for this is mostly identical to that of the 'acx' container, only differing in endianness and magic number. Should their source files be merged to reduce duplication?

### DIFF
--- a/src/libvgmstream.vcxproj
+++ b/src/libvgmstream.vcxproj
@@ -440,6 +440,7 @@
     <ClCompile Include="meta\adx.c" />
     <ClCompile Include="meta\adx_monster.c" />
     <ClCompile Include="meta\afc.c" />
+    <ClCompile Include="meta\afs.c" />
     <ClCompile Include="meta\agsc.c" />
     <ClCompile Include="meta\ahv.c" />
     <ClCompile Include="meta\ahx.c" />
--- a/src/libvgmstream.vcxproj.filters
+++ b/src/libvgmstream.vcxproj.filters
@@ -1150,6 +1150,9 @@
     <ClCompile Include="meta\afc.c">
       <Filter>meta\Source Files</Filter>
     </ClCompile>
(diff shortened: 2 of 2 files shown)


bnnm:
Not sure if this format is a good fit for vgmstream, so let me think a bit first.

AFS is a generic archive file system, which may include anything else like textures or models, separate or together with audio. It often includes filenames and is well known so many tools handle it ([example](https://aluigi.altervista.org/bms/afs.bms)) and most people wouldn't think of vgmstream at this point.

Internal files may also be .AHX (IIRC .ACX too), so the parser fails for various .AFS (incomplete = bug reports, etc).

AFS mainly evolved into other of CRI's filesystem like .CVS/ROFS and .CPK (generic containers), unlike .ACX or .AWB (AFS2) which are used for audio only ([audio banks](https://github.com/vgmstream/vgmstream/blob/master/doc/USAGE.md#containers)).

Since we need to draw the line somewhere, maybe just making a [TXTH with subfiles](https://github.com/vgmstream/vgmstream/blob/master/doc/TXTH.md#grandia-ps1-bgmtxth) is preferable if one really needs to play certain .AFS w/ ADX as-is.

bnnm:
I've decided to reject this based on the points above. I appreciate the PR, but feel that it strays a bit too much from what vgmstream should handle.

As explained if you check those .AFS they have filenames such as BGM_xx.adx at the bottom. You can also see Sonic Gems Collections has .AFS with regular files (ex. MANUAL folders).

You can make a simple [TXTH](github.com/vgmstream/vgmstream/blob/master/doc/TXTH.md) for this format if needed to keep as-is though:
```
subsong_count = @0x04
subsong_spacing = 0x08

subfile_offset = @0x08
subfile_size = @0x0c
subfile_extension = adx
```

PR #1829, opened 2026-01-04
FSB: use FEV1 stream names

EdnessP:
- Add FEV1 v7.0\~v69.0 parsing (tested with v7.0\~v64.0)
- Prioritise FEV stream names over FSB, if present

FEV stream names are essentially the same as those present in FSB, just not truncated to the first 30 chars of the base name.  Some games like Split/Second benefit greatly from this, where the base names are all essentially the same.
Some things could still be improved, such as adding sound definition names, where one track can be a part of multiple defs, e.g. Split/Second again, where there are 15 race tracks but 10 victory tracks, but they still have definitions for 15 of them, just repeating them (SS01_ResultScreen.wav is a part of `/SS01/ResultScreen` and `/SS11/ResultScreen`, etc.), but otherwise it should work properly.
attels

### DIFF
--- a/src/meta/fsb.c
+++ b/src/meta/fsb.c
@@ -682,7 +682,7 @@ static void get_name(char* buf, fsb_header_t* fsb, STREAMFILE* sf_fsb) {
 
         sf_fev->stream_index = sf_fsb->stream_index;
         if (!parse_fev(&fev, sf_fev, filename))
-            vgm_logi("FEV: Failed to parse FEV1 data");
+            vgm_logi("FSB: Failed to parse FEV1 data");
     }
 
     /* prioritise FEV stream names, usually the same as the FSB name just not truncated */


bnnm:
Some comments:
- it fails when parse_fev doesn't succeed, but being optional that seems unnecessary (if one removes the failing .fev then its .fsb would work); a warning/`vgm_logi` would be better

- since fev parse/SF handling doesn't have much to do with the main `init_vgmstream_fsb`, I think it would be better to isolate it in some kind of `load_name` function

- if you could include in some comment that Split/Second particularly benefits from these names, for later reference (I get it's in the PR but I think it's good that this info lives in "git" rather than in "github")

- no used, but the loop [here](https://github.com/vgmstream/vgmstream/blob/4fd7ace2a8c29f25b0d572df2fedfcf7607a67ec/src/meta/fsb_fev.h#L766) may need to init comp_end to SF size to avoid an infinite loop on bad data

- seems I forgot a [rogue log](https://github.com/vgmstream/vgmstream/blob/4fd7ace2a8c29f25b0d572df2fedfcf7607a67ec/src/meta/fsb.c#L230C13-L230C28), if you could remove it while you are at it

EdnessP:
Done. Somewhat followed the example in XWB+XSB by isolating it into a `get_name` function.

PR #1834, opened 2026-01-09
FSB: RIFF FEV stream names

EdnessP:
Adds RIFF FEV parsing (v65.0+).  Turns out RIFF FEV was also briefly used for FSB4, not just FSB5, and they basically just embedded the same FEV1 data into it.  This does not include the newer .bank with FSB5 embedded into a RIFF FEV, those are its entirely separate thing.

Some streams have multiple names, which are now also displayed.  Another thing of note is I noticed several FSB5 files where some full stream names are present in the original Sound Defs area, but some are only present in the later Composition data chunk even within the same FSB+FEV pair, which is what I'll work on next soon-ish.

In my previous PR I mentioned how I wanted to also append and display the main sound definition names along the real stream names, but that ended up looking way too cluttered and messy (especially after finding out there can be multiple names assigned to the same sound too) so I've since dropped that idea.

I also updated the chunks util to support explicit aligned scans RIFF style, FEV has chunks with strings where this will break if the 16-bit alignment isn't enforced.  (In fact, I only just now noticed as of writing this, the alternate `next_chunk()` parser has an `alignment` flag in `chunks_t`, yet it doesn't seem to be actually used.  Shall I fix that too?)

### DIFF
--- a/src/meta/fsb_fev.h
+++ b/src/meta/fsb_fev.h
@@ -59,14 +59,14 @@ typedef struct {
 #define FMOD_FEV_VERSION_34_0   0x00220000  // Ys Online: The Call of Solum (PC)
 //#define FMOD_FEV_VERSION_35_0 0x00230000  // ?
 #define FMOD_FEV_VERSION_36_0   0x00240000  // ?
-#define FMOD_FEV_VERSION_37_0   0x00250000  // Conan (X360)
+#define FMOD_FEV_VERSION_37_0   0x00250000  // Conan (X360), Manhunt 2 (Wii)
 #define FMOD_FEV_VERSION_38_0   0x00260000  // ?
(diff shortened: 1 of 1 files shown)


bnnm:
Feel free to fix `alignment` if you want (and maybe we should use `bool` for flags), kind of forgot about it since most pseudo-riffs don't follow that rule.

Also it's fine to keep it as is, but as for alt parsers `read_fev_x` seem equivalent to `read_helper.h` fns too. We can also define variables mid-code (if there isn't a `goto` before) if you prefer.

EdnessP:
Nice, didn't know about reader_helper, replaced my FEV reimpl with that.  But personally I still prefer variable declarations at the top of a scope.

Also last minute fix I just noticed, Manhunt 2 (Wii/PC) bank names include a directory, which has to be stripped for it to correctly find and assign the right bank.

PR #1844, opened 2026-01-20
Add omt/0MF2

Alstruit:
- AWE Productions
   - SpongeBob SquarePants: Employee of the Month
   
   [Samples](https://0x0.st/s/XUiGN7whKvXA7BlD1w2BnA/PKYp.zip)
   [Logs](https://0x0.st/s/1qVOjhUgqKu2USvAsg343w/PKY_.zip)

### DIFF
--- a/src/formats.c
+++ b/src/formats.c
@@ -433,6 +433,7 @@ static const char* extension_list[] = {
     "ogv",
     "oma", //FFmpeg/not parsed (ATRAC3/ATRAC3PLUS/MP3/LPCM/WMA)
     "omu",
+    "omt",
     "oor",
     "opu",
     //"opus", //common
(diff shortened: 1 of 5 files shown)


bnnm:
I don't think this is a sound format, and instead just a generic container. For example, it's also used in 'scripts' to contain non-audio data. There is a OMT.dll that handles multiple things unrelated to audio.

It also would be pretty strange to have a sound format that needs to manually scan bytes to detect "wave" sections. Games don't have CPU to spare so almost always their formats have a normal structure. This seems to me a more logical logical way to parse .omt: [omf2.zip](https://github.com/user-attachments/files/24782346/omf2.zip) (quickbms script)

Since generic containers are typically outside vgmstream's realm, unless there is a good reason I don't think this is a good fit ([see here](https://github.com/vgmstream/vgmstream/pull/1812#issuecomment-3620464063) for a similar reasoning).

Alstruit:
> I don't think this is a sound format, and instead just a generic container. For example, it's also used in 'scripts' to contain non-audio data. There is a OMT.dll that handles multiple things unrelated to audio.
> 
> It also would be pretty strange to have a sound format that needs to manually scan bytes to detect "wave" sections. Games don't have CPU to spare so almost always their formats have a normal structure. This seems to me a more logical logical way to parse .omt: [omf2.zip](https://github.com/user-attachments/files/24782346/omf2.zip) (quickbms script)
> 
> Since generic containers are typically outside vgmstream's realm, unless there is a good reason I don't think this is a good fit ([see here](https://github.com/vgmstream/vgmstream/pull/1812#issuecomment-3620464063) for a similar reasoning).

Thanks for the clarification. I suppose the Wave check is insufficent since that clearly marks audo files only.

PR #1846, opened 2026-01-20
Add ps2p from THQ Australia

Alstruit:
- [Samples](https://0x0.st/s/bQNxHKk--zhDMMZ_2srFxA/PKI5.zip)
- [Logs](https://0x0.st/s/OwCr7P13DGaecoea7zJUVQ/PKIn.zip)

### DIFF
--- a/src/meta/ps2p.c
+++ b/src/meta/ps2p.c
@@ -8,7 +8,7 @@
 #define PS2P_ENTRY_SIZE         0x0C
 #define PS2P_AUX_ENTRY_SIZE     0x1C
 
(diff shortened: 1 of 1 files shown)


Alstruit:
[Updated logs](https://0x0.st/s/2cp0sy11e7__giGJmSHj-w/PPTE.zip)

bnnm:
Since this has a clear fourcc, I think file makes more sense to be named `ps2p.c` and define `init_vgmstream_ps2p` and `meta_PS2P`. 

Company name is typically only appended in special cases such as complex sound engines.

Alstruit:
Well there is the gamecube variant that might have a similar table structure. I haven't gone ahead and implement those yet. Xbox variant does not have the fourcc but appears to have XWB/WBND. I've made the changes.

PR #1904, opened 2026-04-08
Increase XA interleaved channels limit to 128

N4gtan:
This PR increases the `XA_MAX_CHANNELS` constant from 32 to 255.

**Context:**
While most PS1 titles use between 8 and 16 interleaved channels, some specific games like _Digimon Rumble Arena_ use up to ~120 channels. The current limit of 32 causes these files to be ignored by the metadata parser.

**Technical Choice:**
The limit is set to 255 instead of 256 to maintain the existing validation logic. Since `xa_chan` is a `uint8_t`, this change allows the parser to process all channels up to 254 while still treating `0xFF` (255) as an invalid/null channel. This is important for games like _Yu-Gi-Oh! Forbidden Memories_, _Ace Combat_, etc., which use `0xFF` for padding/silent sectors.

### DIFF
--- a/src/meta/xa.c
+++ b/src/meta/xa.c
@@ -216,7 +216,10 @@ fail:
 }
 
 
-#define XA_MAX_CHANNELS 32 /* usually 08-16, seen ~24 in Langrisser V (PS1) */
+/* usually 08-16; Digimon Rumble Arena uses ~120. We use 255 to support many
+ * interleaved channels while excluding 0xFF, as it is commonly used for
+ * null/silence sectors in games like Yu-Gi-Oh! FM, Ace Combat, etc. */
+#define XA_MAX_CHANNELS 255
 
 typedef struct {
    uint32_t info;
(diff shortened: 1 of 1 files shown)


bnnm:
Setting it to 255 would still fail in those padding sectors, since it's doing `if (xa_chan >= XA_MAX_CHANNELS) goto fail;` (0xFF == 255).

Looking at Yu-Gi-Oh, it seems like a special case that should be ignored so it's better to do add something like this before the above check:
```
        // padding/silent sectors [Yu-Gi-Oh! Forbidden Memories (PS1)]
        if (xa_file == 0xFF && xa_chan == 0xFF) {
            offset += sector_size;
            continue;
        }
```
XA_MAX_CHANNELS can set to around 128 (highest known). It's mainly there to catch odd behaviors like this one.

For Digimon Rumble Arena the extension used seems to be `.xap` so it may be useful to add to `check_extensions` too.

N4gtan:
> Setting it to 255 would still fail in those padding sectors, since it's doing `if (xa_chan >= XA_MAX_CHANNELS) goto fail;` (0xFF == 255).

Right, actually I didn't noticed because I was testing with already deinterleaved files.
But still the check `if (xa_file == 0xFF && xa_chan == 0xFF)` will fail, as the game filenum value is 0x08, not 0xFF. I tested it with the Spain version, maybe the USA version has a filenum 0xFF?
I think it would be better to only check for `xa_chan` value.

bnnm:
In the US version xa_file is always 0xFF for padding sectors, but if other versions only use xa_chan that's fine too.

PR #1937, opened 2026-06-20
Add CyberFlix DreamFactory D5

Alstruit:
Adds support for CyberFlix DreamFactory v5.x.
Titles: Redjack Revenge of the Brethern, Disney's Villain's Revenge, Disney's The D show

[Samples](https://files.catbox.moe/tg7dda.zip)
[Logs](https://files.catbox.moe/mll8sb.zip)


bnnm:
Please tweak the code so that `goto` can't go over variable definitions outside a block. Example:
```
    ...
    if (!soun_ids) goto fail;

    int soun_count = 0;   //compiler warning, declaration right after goto outside a block
    ...
```
While they may not matter much in this case they are valid warnings for other cases.

Issue #1999, opened 2026-08-20
[Bug] Divide by zero in Microsoft IMA ADPCM block decoding


ni-liao:
### Summary
A crafted RIFF/WAVE file can trigger a divide-by-zero in Microsoft IMA ADPCM block decoding, causing denial of service in applications embedding vgmstream.

### Details
I re-checked the latest upstream HEAD (`beb753e6ddeeff72143e5d2750760f97260c8fe4`) and this issue still appears to be present.

The current `decode_ms_ima()` implementation still derives:

- `frame_channels = vgmstream->codec_config ? 1 : vgmstream->channels`
- `block_samples = ((frame_size - 0x04 * frame_channels) * 2 / frame_channels) + 1`
- `first_sample = first_sample % block_samples`

There is still no guard for `frame_channels <= 0` and no guard for `block_samples <= 0` before the modulo operation.

A malformed header can therefore still force invalid decoder geometry into the arithmetic path.

### Reproduction
I have a minimized local testcase and prior reproduction artifacts for this root cause. I can provide the testcase, logs, and exact replay steps if needed.

### Impact
This is a denial-of-service issue in the decoder path. Any application that opens untrusted RIFF/WAVE content through vgmstream can be terminated during MS IMA decoding.

### Suggested fix
Reject non-positive channel geometry before computing the derived block size, and return early if the derived `block_samples` value is non-positive.

Issue #1998, opened 2026-08-20
[Bug] Use after free in TXTP random-group failure cleanup


ni-liao:
### Summary
A crafted TXTP file can trigger a second close of a freed `VGMSTREAM` object during random-group failure cleanup, causing process termination and potentially enabling further memory corruption.

### Details
I re-checked the latest upstream HEAD (`beb753e6ddeeff72143e5d2750760f97260c8fe4`) and this issue still appears to be present.

The current `make_group_random()` logic still:

- selects a `vgmstream` into a local pointer
- updates the shared list with `update_vgmstream_list(vgmstream, ...)`
- later calls `make_group_segment(..., 1)` under the config path
- and on failure still falls through to `close_vgmstream(vgmstream)`

This means the caller can still close a stale pointer after ownership has already been transferred or freed by the failed segment-construction path.

### Reproduction
I have a minimized local testcase and prior ASan reproduction artifacts for this root cause. I can provide the testcase, logs, and exact replay steps if needed.

### Impact
This is a use-after-free in cleanup logic. The immediate result is a crash, but because the stale pointer is reused in object teardown, the bug should be treated as memory corruption rather than a pure denial of service.

### Suggested fix
Clear the caller's local pointer after a failed ownership transfer, or restructure the control flow so that the shared failure path never closes an object that may already have been consumed by `make_group_segment()`.

Issue #1997, opened 2026-08-20
[Bug] Heap buffer overflow in render silence padding


ni-liao:
### Summary
A crafted TXTP file can drive an out-of-bounds write into the render buffer during silence padding, causing process termination and potentially enabling memory corruption in applications embedding vgmstream.

### Details
I re-checked the latest upstream HEAD (`beb753e6ddeeff72143e5d2750760f97260c8fe4`) and this issue still appears to be present.

The vulnerable helper `sbuf_silence_part()` is still unchanged on HEAD. It still:

- computes `buf += from * sbuf->channels * sample_size`
- calls `memset(buf, 0, count * sbuf->channels * sample_size)`

Those calculations still use signed integers and there are still no range checks on `from`, `count`, or the resulting byte span before the write.

This leaves the original root cause intact when attacker-controlled TXTP timing reaches this helper with an invalid range.

### Reproduction
I have a minimized local testcase and prior ASan reproduction artifacts for this root cause. I can provide the testcase, logs, and exact replay steps if needed.

### Impact
This is a heap memory corruption issue in the render path. A crafted input can write beyond the heap-backed sample buffer and crash the host process; depending on embedding context, exploitation beyond a crash may be possible.

### Suggested fix
Validate that `from` and `count` stay within the sample-buffer capacity, reject negative values, and perform byte-size calculations in `size_t` after bounds validation.

PR #2002, opened 2026-08-20
fix: guard against divide-by-zero, heap overflow, and use-after-free in crafted file paths (Fixes #1996, #1997, #1998, #1999)

### LINKED ISSUE #1996: [Bug] Divide by zero in AWB offset alignment parsing
### Summary
A crafted AWB file can trigger a divide-by-zero during offset-table parsing, causing denial of service in applications embedding vgmstream.

### Details
I re-checked the latest upstream HEAD (`beb753e6ddeeff72143e5d2750760f97260c8fe4`) and this issue still appears to be present.

The current `init_vgmstream_awb_memory()` implementation still reads the attacker-controlled 16-bit `offset_alignment` field and then uses it directly in these modulo expressions:

- `subfile_offset % offset_alignment`
- `subfile_next % offset_alignment`

There is still no explicit `offset_alignment == 0` validation before those operations.

### Reproduction
I have a minimized local testcase and prior reproduction artifacts for this root cause. I can provide the testcase, logs, and exact replay steps if needed.

### Impact
This is a denial-of-service issue. Any application that opens an untrusted AWB/AFS2/AWX file through vgmstream can be terminated during metadata parsing before audio decoding begins.

### Suggested fix
Reject zero alignment before parsing the offset table and keep the existing alignment logic only for strictly positive values.

waterWang:
## Summary

Four security fixes for crafted file handling, all against the same HEAD (`beb753e`).

### Fixes

**#1996** — AWB: check `offset_alignment == 0` before using it in modulo operations.  
A crafted AWB file with `offset_alignment == 0` causes division by zero in offset-table alignment.

**#1997** — sbuf: add bounds check in `sbuf_silence_part()` before writing to the buffer.  
A crafted TXTP file can drive `from` or `count` past `sbuf->samples`, causing heap buffer overflow.

**#1998** — txtp: nullify the selected vgmstream pointer after `make_group_segment()` takes ownership.  
When `make_group_segment()` fails inside `make_group_random()`, the segment layout's `free_layout_segmented` already closes the sub-streams, but the caller's fail path calls `close_vgmstream()` again on the same pointer → use-after-free / double-free.

**#1999** — ima: guard `frame_channels` and `block_samples` against zero in `decode_ms_ima()`.  
A crafted MS IMA ADPCM stream with `channels == 0` or a malformed frame_size causes division by zero.

### Testing

- Builds successfully with `make -j$(nproc)` (no new warnings).
- Each fix is a defensive guard that returns early or skips the dangerous operation.

### DIFF
--- a/src/base/sbuf.c
+++ b/src/base/sbuf.c
@@ -206,10 +206,20 @@ int sbuf_get_copy_max(sbuf_t* sdst, sbuf_t* ssrc) {
 
 void sbuf_silence_part(sbuf_t* sbuf, int from, int count) {
     int sample_size = sfmt_get_sample_size(sbuf->fmt);
+    size_t byte_count;
+
+    /* check span before computing the pointer (from/count come from untrusted config) */
+    if (from < 0 || count < 0 || from > sbuf->samples || count > sbuf->samples - from) {
+        VGM_LOG("sbuf: silence over buffer (%i+%i > %i)\n", from, count, sbuf->samples);
+        return;
+    }
+
+    /* check overflow of the byte count (only for safety, spans are checked above) */
+    byte_count = (size_t)count * (size_t)sbuf->channels * (size_t)sample_size;
 
     uint8_t* buf = sbuf->buf;
-    buf += from * sbuf->channels * sample_size;
-    memset(buf, 0, count * sbuf->channels * sample_size);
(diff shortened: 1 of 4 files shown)


bnnm:
This felt too LLM and not proper fixes so closing it.

